PR #6, opened 2014-07-02
Do not check flake8 in __unported__

dreispt:
And get TravisCI green!


dreispt:
Won't have any effect here, since there is no `__unported__` dir, but it worked [here](https://github.com/OCA/department/pull/6) and this repo will be the reference for setting up other projects.

pedrobaeza:
:+1:

guewen:
:+1: 

Thanks!

PR #13, opened 2014-08-05
[IMP] add the posibility to merge the job. This is very usefull for the ...

sebastienbeau:
...Magento case, indeed the api is very slow and we have to avoid the useless call to Magento, if we modify a binding and a product only one job should be created in order to avoid 2 call to Magento instead of one

You can take a look to this exemple : https://github.com/OCA/connector-magento/pull/23

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -103,16 +103,21 @@ class OpenERPJobStorage(JobStorage):
             "Model %s not found" % self._job_model_name)
 
     def enqueue(self, func, model_name=None, args=None, kwargs=None,
-                priority=None, eta=None, max_retries=None, description=None):
+                priority=None, eta=None, max_retries=None, description=None,
+                _job_to_key=None, _merge_job=None):
         """Create a Job and enqueue it in the queue. Return the job uuid.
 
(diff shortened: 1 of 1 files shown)


guewen:
Discussed off-topic, needs to be done in another manner, with a `mergeable` declaration on the jobs using a decorator. Optionally, a function may be given that explicit how the kwargs are merged.

nbessi:
May we close this PR and make a new one when ready?

rvalyi:
great, I want it too.

lmignon:
Can we close this PR? It seems to be obsolete and there is no more activity?

hbrunn:
@lmignon after three months, I'd simply do so if I were you. Or don't you have permissions?

lmignon:
@hbrunn I've the permission but I prefer to ask by courtesy to the author ho has invested to enhance the connector :smirk:

hbrunn:
@lmignon being a python programmer, I think it's easier to ask for forgiveness than permission ;-)

sebastienbeau:
For sure you can close it. As I am not available to work on it for now

PR #19, opened 2014-09-10
Fixes #18. The init of `Environment` object is broken 

### LINKED ISSUE #18: In version 8.0 getting connector unit breaks with last version of API

nbessi:
as backend model are now new Odoo API model.

The result  of function `get_backend` is mapped to a list so we have a list of `backend.Backend` object (not model)
It seems it is possible to have correct result using decorator and downgrade keywork but I do not like it.

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -236,7 +236,8 @@ class Environment(object):
         :type model_name: str
         """
         self.backend_record = backend_record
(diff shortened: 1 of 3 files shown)


nbessi:
Maybe a better option is to make the function private _get_backend

lmignon:
:+1:

guewen:
:+1:

PR #24, opened 2014-10-16
7.0 workaround oca issue 22

tremlin:
This attempts to work around OCA issue 22, "AssertionError: 0 worker found in database instead of 1". After I encountered this problem in a second installation, I decided to create a pull request. I don't know what causes the assertion to fail (probably some timing issue or a problem with starting Odoo and the connector with too few workers), but it makes the connector work again for the two installations.

### DIFF
--- a/connector/queue/model.py
+++ b/connector/queue/model.py
@@ -339,7 +339,11 @@ class QueueWorker(orm.Model):
             return
         job_ids = [id for id, in job_rows]
 
-        worker_id = self._worker_id(cr, uid, context=context)
+        try:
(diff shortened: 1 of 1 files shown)


guewen:
I wrote a similar fix (then reverted it) some time ago if I remember well, I finally fixed the original error in the odoo core which was a race condition in the Registry (maybe it is an other occurrence of this bug or maybe that's just a normal behavior). Anyway, I agree with this check. :+1: 

Thanks!

lmignon:
:+1:

PR #25, opened 2014-10-20
[IMP] Call 'only_create' mapping functions only when the mapper is calle...

lmignon:
...d  with for_create=True

Some mapping functions can be resource intensive. The change avoids calling functions annotated with 'only_create' if the mapper is not called with for_create=True and thus reduces resource consumption in such cases.


guewen:
It makes sense indeed!
Thanks 
:+1:

pedrobaeza:
:+1:

PR #26, opened 2014-10-20
[FIX] When determining the name of the field mapped by a function in the...

lmignon:
... direct mappings, the argument order seems to be not enforced by python in the closure so we use the first non callable cell_contents in the closure as attr_name

Some times, tests fails on my machine due to a bug fixed in 8.0 but not yet back-ported. Fix back-ported by cherry-pick of https://github.com/OCA/connector/commit/e16533b092515d952d73ffd4ff2ee36573d2d64e


guewen:
:+1: 
Thanks

guewen:
Only my thumb up, but it as been accepted on 8.0.

PR #29, opened 2014-11-05
name_get implementations expect a list of ids

mgrohmann:
the original orm.py name_get function checks the type of 'ids' but not so most of the addons implementations


pedrobaeza:
Good safeguarding!

:+1:

guewen:
:+1: 
Thanks for your proposal!

PR #30, opened 2014-11-14
[REF] auto-fix conventions

moylop260:


guewen:
Disagree because it adds blank lines between the class definition and the docstring. :-1: 
This is not what says the pep0257: http://legacy.python.org/dev/peps/pep-0257/

guewen:
Thanks for your PR, I corrected the remaining warnings and removed the lines on #33

PR #31, opened 2014-11-19
[FIX][connector] confusion between self.env from test of connector and s...

JonathanNEMRY:
...elf.env recently added into tearDown of common.py

see https://github.com/odoo/odoo/commit/908252ec88cd4516ee84fdf93c07d09f51a45413

PR #32, opened 2014-11-19
[FIX][connector] switch unlink of producer into the new api to avoid old...

JonathanNEMRY:
.../new odoo confusion
- Create a Database
- Install "Products & Pricelists"  & "Purchase" & "Sales Management"
- Settings/Configuration/Purchases --> Manage different units of measure for products
- Settings/Configuration/Purchases --> Manage pricelist per supplier
- Invoicing/Create customer "customer"
- Édit "customer" and change its "Purchase Pricelist" (create a second one before)
- Install connector module
- Edit the created "customer" and change its "Purchase Pricelist" with the default one (it must be the real default purchase pricelist because it will make an unlink)
- Save and see "TypeError: unlink() takes at least 4 arguments (1 given)"
## Into 

```
`@api.model
 def set_multi(self, name, model, values):`
```

of `ir.property`
## Behavior without connector:

the `prop.unlink()` is well redirected to 
    `BaseModel.unlink(cr, uid, ids, context)` passing by the wrapper/new_api

**--> works pretty well**

---
## Behavior with connector:

the `prop.unlink()` is NOT well redirected:

```
`producer.unlink(cr, uid, ids, context)` passing by the wrapper/OLD_API
```

**--> takes at least 4 arguments (1 given)**
## Supposion:

I personally do not think this is a `connector` bug because it mostly appears as a 
`old/new api` problem (seems that `unlink` is called as new `api` from `ir.property`
but not defined as such into `BaseModel`).

**BUT** it can be resolved by switching the `producer.unlink` into new `api`.

If you have some other solution and/or any ideas

(Note that test will fail:  maybe rebase after merging this https://github.com/OCA/connector/pull/31)

-jne

### DIFF
--- a/connector/producer.py
+++ b/connector/producer.py
@@ -63,7 +63,7 @@ def write(self, vals):
         session = ConnectorSession(self.env.cr, self.env.uid,
                                    context=self.env.context)
         if on_record_write.has_consumer_for(session, self._name):
-            for record_id in self._ids:
+            for record_id in self.ids:
                 on_record_write.fire(session, self._name,
                                      record_id, vals)
     return result
@@ -79,7 +79,7 @@ def unlink(self):
         session = ConnectorSession(self.env.cr, self.env.uid,
                                    context=self.env.context)
         if on_record_unlink.has_consumer_for(session, self._name):
-            for record_id in self._ids:
+            for record_id in self.ids:
                 on_record_unlink.fire(session, self._name, record_id)
     return unlink_original(self)
 orm.BaseModel.unlink = unlink
--- a/connector/producer.py
+++ b/connector/producer.py
@@ -73,13 +73,13 @@ orm.BaseModel.write = write
 unlink_original = orm.BaseModel.unlink
 
 
-def unlink(self, cr, uid, ids, context=None):
+@openerp.api.multi
+def unlink(self):
     if self.pool.get('connector.installed') is not None:
-        if not hasattr(ids, '__iter__'):
-            ids = [ids]
-        session = ConnectorSession(cr, uid, context=context)
+        session = ConnectorSession(self.env.cr, self.env.uid,
+                                   context=self.env.context)
         if on_record_unlink.has_consumer_for(session, self._name):
-            for record_id in ids:
+            for record_id in self._ids:
                 on_record_unlink.fire(session, self._name, record_id)
-    return unlink_original(self, cr, uid, ids, context=context)
+    return unlink_original(self)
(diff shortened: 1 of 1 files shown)


connector/producer.py
```diff
@@ -73,13 +73,13 @@ def write(self, vals):
...
+def unlink(self):
     if self.pool.get('connector.installed') is not None:
-        if not hasattr(ids, '__iter__'):
-            ids = [ids]
-        session = ConnectorSession(cr, uid, context=context)
+        session = ConnectorSession(self.env.cr, self.env.uid,
+                                   context=self.env.context)
         if on_record_unlink.has_consumer_for(session, self._name):
-            for record_id in ids:
+            for record_id in self._ids:
```
guewen: could you use `self.ids`?

guewen:
Thanks! Seems fine, just one little thing with the `self.ids`.

lmignon:
:+1:  once the the self._ids will be fixed :smile:

AdilHoumadi:
:+1:

pedrobaeza:
:+1:

ghost:
:+1:

guewen:
Just waiting after travis...

guewen:
:+1:

pedrobaeza:
UNIT_TEST is provoking this slowness...

guewen:
I think that we have a record of thumbs up per hour in the OCA on this PR.

pedrobaeza:
:laughing:

JonathanNEMRY:
Rebase after merge of #31

guewen:
I don't think the rebase was necessary because travis run the test with the branch merged on the HEAD of the base branch.

JonathanNEMRY:
@guewen  thanks for this information

guewen:
Should be green after merge of https://github.com/OCA/connector/pull/34
Anyway the failure is not related so I proceed with the merge.

Thanks!

lepistone:
Hi all,

after some investigation I found out what was going on here.

The original `unlink` method is not decorated, and written with the old API. This is OK, because the `api.Meta` metaclass is used in the `BaseModel` and guesses the decorator automatically when it is not provided. In that case, the connector replaces the `unlink` method with a new, non decorated one, and the guessing never takes place (because the method is not defined in `BaseModel`).

This PR fixed the problem because the decorator `@api.multi` has been specified. I think just adding the old-style `@api.cr_uid_ids_context` would have worked as well.

Thanks to you all for your time!

PR #46, opened 2015-02-19
backend_record.get_backend *must* return a backend, not a list of backends

guewen:
The previous fix took the first item of the list on the result of the method,
but it adds the burden on all the callers to do the same thing.

Instead, changing the 'id' to 'ids' in the method's arguments makes the API use
the correct wrapper and returns one backend instead of a list.

Reverts f5bf250 and 6bf7299


pedrobaeza:
:+1:

lmignon:
:+1:

yvaucher:
:+1:

PR #47, opened 2015-02-19
New API in connector, bump to 3.0.0

guewen:
The changes done here will break the API compatibility.
However, they are necessary to keep up with the changes in Odoo.

The API in the connector still uses the `cr, uid, ids, context` api, even if
hidden in `connector.Session`. Not using new recordsets is restricting.

Example:
If I implements a new onchange on `sale.order`:

``` python
class SaleOrder(models.Model):
    @api.onchange('workflow_process_id'):
    def onchange_workflow_process_id(self):
        self.payment_method_id = self.workflow_process_id.payment_method_id
```

Then I won't be able to call from within a `ConnectorUnit` directly.
Locally I could do:

``` python
import openerp
[...]

  env = openerp.api.Environment(self.session.cr,
                                self.session.uid,
                                self.session.context)
  sale = env['sale.order'].new(values)
  sale.onchange_workflow_process_id()
```

But it would be far better to include the new API in the root classes.
# Changes:
- Add `openerp.api.Environment` in `Session`
  It is accessible in `self.env` in `Session` and all
  `ConnectorUnit` instances.
  Also in `ConnectorUnit`, `model` returns the current (new api!) model:
  
  ``` python
      # On the current model
      self.model.search([])
      self.model.browse(ids)
      # on another model
      self.env['res.users'].search([])
      self.env['res.users'].browse(ids)
  ```
- Deprecate the CRUD methods in `Session`
  
  ``` python
      # NO
      self.session.search('res.partner', [])
      self.session.browse('res.partner', ids)
  
      # YES
      self.env['res.partner'].search([])
      self.env['res.partner'].browse(ids)
  ```
- `Environment.set_lang()` is removed. It was modifying the context
  in place which is not possible with the new frozendict context. It
  should be done with:
  
  ``` python
  
      with self.session.change_context(lang=lang_code):
          ...
  ```
- Add an argument on the Binders methods to return a browse record
  
  ``` python
  
      binder.to_openerp(magento_id, browse=True)
  ```
- Shorten `ConnectorUnit.get_binder_for_model` to
  `ConnectorUnit.binder_for`
- Shorten `ConnectorUnit.get_connector_unit_for_model` to
  `ConnectorUnit.unit_for`
- Renamed `Environment` to `ConnectorEnvironment` to avoid
  confusion with `openerp.api.Environment`
- Renamed the class attribute `ConnectorUnit.model_name` to
  `ConnectorUnit.for_model_name`.
- Added `_base_binder`, `_base_mapper`, `_base_backend_adapter` in
  the synchronizers (Importer, Exporter) so it is no longer required to
  override the `binder`, `mapper`, `backend_adapter` property
  methods
- `Session.change_context()` now supports the same
  argument/keyword arguments semantics than
  `openerp.model.BaseModel.with_context()`.
- Renamed `ExportSynchronizer` to `Exporter`
- Renamed `ImportSynchronizer` to `Importer`
- Renamed `DeleteSynchronizer` to `Deleter`
- `Session.commit` do not commit when tests are running
- Cleaned the methods that have been deprecated in version 2.x


connector/session.py
```diff
@@ -102,33 +110,37 @@ class ConnectorSession(object):
...
-        return self._pool
+        _logger.warning("Deprecated: use 'registry'")
+        return self.registry
+
+    @contextmanager
+    def change_user(self, uid):
+        """ Context Manager: temporarily change the user's session and
+        restablish the normal user at closing,
+        """
+        raise Exception('Deprecated: use sudo() on a recordset.')
```
guewen: Maybe just yield a new Environment, and restore the original at closing.

connector/session.py
```diff
@@ -138,11 +150,7 @@ def change_context(self, values):
         :param values: values to apply on the context
         :type values: dict
         """
-        original_context = self.context
-        self._context = original_context.copy()
-        self._context.update(values)
-        yield
-        self._context = original_context
+        raise Exception('Deprecated: use with_context() on a recordset.')
```
guewen: Maybe just yield a new Environment, and restore the original at closing.

connector/connector.py
```diff
@@ -166,6 +170,22 @@ def match(cls, session, model):
...
+
+    @property
+    def model(self):
+        _logger.warning('Deprecated: ConnectorUnit.model is no longer a '
+                        'property. Now, you should use '
+                        'ConnectorUnit.recordset() that returns a '
+                        'new-api model/recordset.')
+        return self.session.pool.get(self.environment.model_name)
+
+    def recordset(self, model=None):
```
guewen: The perfect name for this method would have been `model`... but it is already used.
`recordset` is a bit odd as it returns a model (but a model is also a new recordset in the new api), but I can't find a better name.

guewen: What about `records`?

``` python
users = self.records().search([])  # when the model of the ConnectorUnit is 'res.users'
users = self.records('res.users').search([])  # when the model is another thing
things = self.records().browse(ids)
things.unlink()
```

guewen:
Branch's ready, I need reviews

lmignon:
@Thank you @guewen for this hard work... Even if there are a lot of changes, I did not find any comments to make. :+1:  (Code review)

pedrobaeza:
Well, I have to check it when I change Prestashop connector, but at first glance, it seems :+1:

I think we can merge and then make possibles PRs to fix problems.

tremlin:
I reviewed the code and tested it (as part of `magentoerpconnect`), I recommend it for merging :+1: 
Thanks for the port to v8!

pedrobaeza:
I proceed with the merge then and possible fixes will come as another PRs.

Thanks to all that have contributed!

pedrobaeza:
@guewen, can you please include the same version log of this PR in any side of the web to have it as reference?

connector/CHANGES.rst
```diff
@@ -1,6 +1,74 @@
 Changelog
 ---------
 
+3.0.0 (not released)
```
yvaucher, line 4: Date needs to be set

pedrobaeza, line 4: I think you can make it via direct commit.

guewen, line 4: done

guewen:
> @guewen, can you please include the same version log of this PR in any side of the web to have it as reference?

What do you mean exactly by "in any side of the web"?
I updated http://odoo-connector.com if that's what you mean.

pedrobaeza:
Yeah, http://odoo-connector.com/project/changes.html#id2 is what I mean. Thanks.

PR #50, opened 2015-03-13
Make the workers process work in a rolling order

guewen:
Assign an index to each worker, then at each turn, add a delay relative to
their index modulo the population.

This way, each worker has an equal opportunity to take jobs.

Example with 4 workers
- loop 1
  - worker 1: sleep 15 + 1
  - worker 2: sleep 15 + 2
  - worker 3: sleep 15 + 3
  - worker 4: sleep 15 + 0
- loop 2
  - worker 1: sleep 15 + 2
  - worker 2: sleep 15 + 3
  - worker 3: sleep 15 + 0
  - worker 4: sleep 15 + 1
- loop 3
  - worker 1: sleep 15 + 3
  - worker 2: sleep 15 + 0
  - worker 3: sleep 15 + 1
  - worker 4: sleep 15 + 2
- loop 4
  - worker 1: sleep 15 + 0
  - worker 2: sleep 15 + 1
  - worker 3: sleep 15 + 2
  - worker 4: sleep 15 + 3

### DIFF
--- a/connector/openerp-connector-worker
+++ b/connector/openerp-connector-worker
@@ -79,7 +79,7 @@ class WorkerConnector(workers.Worker):
             if config['db_name']:
                 db_names = config['db_name'].split(',')
             else:
-               db_names = openerp.service.db.exp_list(True)
+                db_names = openerp.service.db.exp_list(True)
             dbfilter = config['dbfilter']
             if dbfilter and db_names:
                 db_names = [d for d in db_names if re.match(dbfilter, d)]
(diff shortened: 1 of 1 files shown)


guewen:
@lmignon looking forward to have your opinion and tests if possible

lmignon:
@guewen  thank you for the proposal.  I'll make a try on monday and keep you informed.

yvaucher:
@lmignon any comment some mondays later ? :wink: 

:+1: (code review)

lmignon:
@guewen @yvaucher Sorry, this PR is fallen in the limbo of nowhere... IMO with the new runner, the change is no more required

guewen:
It would only improve the old poor workers and not the new shiny runner, but maybe we can give some cuddle to the old workers before the day we'll savagely kill them.

Nah, seriously if people suffer from the problem this PR aims to reduce, they just have to use the new runner.

PR #51, opened 2015-03-29
[FIX] connector worker: jobs must be set done in same transaction as job execution

sbidoul:
Otherwise a crash at the wrong moment could leave an incoherent job state.

### DIFF
--- a/connector/queue/worker.py
+++ b/connector/queue/worker.py
@@ -122,11 +122,9 @@ class Worker(threading.Thread):
             _logger.debug('%s started', job)
             with session_hdl.session() as session:
                 job.perform(session)
-            _logger.debug('%s done', job)
-
-            with session_hdl.session() as session:
                 job.set_done()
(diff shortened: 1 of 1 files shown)


guewen:
Thanks for the patch.

I think this was on purpose not to make the transaction inside the job fail if the transaction that changes the state of the job fail. But indeed, the issue is that if it happens, the job has been executed but rest in 'running'.

If the state change makes the transaction fail, I see 2 figures:
1. the job is idempotent / can be rollbacked
2. the job is not idempotent / can not be rollbacked entirely (e.g. export of a record on an external system)

For 1., there is no issue since the job can be retried later. For 2., I think that we can agree that it is the responsibility of the job to ensure that is idempotent or that it commits when it needs absolutely to keep data (e.g. ID of a record just created on a external system).

So, :+1:

yvaucher:
:+1:

Issue #43, opened 2015-01-21
[Draft] Job channels


guewen:
# Job channels
## Introduction

This is a document explaining how would be implemented the channels in the jobs queue.  It is in draft and will be modified.  At time of writing, the _what_ and _how_ regarding the implementation has not been decided yet so read it with caution.
## What they resolve

In the jobs queue, jobs are all pushed in a queue.  Then, one or many processes take them from the queue and run them one after the other.

This has serious limitations in real use cases:
- If one system does not support to be hit by several workers at the same time (Magento, look at me), we will be limited to use one worker even if we have several connectors
- If tasks are very long to execute, they could clog the queue and prevent other little jobs to be executed

This is were the channels come in handy.
- Let's say this very heavy, long job clutters the workers, put it in a channel 'slow' and place only 1 worker on it, keep the other workers execute the rest
- Or we have this system that want to be touched gently, start only 1 worker that will handle all the channels of this system, and 3 other workers for the other connectors
### Assignation of a channel to a job

All jobs that have no channel specified goes into a `default` channel.

A channel can be specified on using the keyword argument `channel` of the `job` decorator:

``` python
@job(channel='channel`)
def import_record(session, model_name, backend_id, external_id):
    pass
```

Sub-channels allow more fine-grain control on the execution of the jobs by the workers.

``` python
@job(channel='channel.subchannel.subsubchannel`)
def import_record(session, model_name, backend_id, external_id):
    pass
```
### Workers

If nothing is specified when the workers are started, the workers watch all the channels.

The usual command starts 2 workers that will executes all the jobs whatever their channel is. So it works in a backward compatible manner:

``` bash
$ PYTHONPATH=/path/to/server connector/openerp-connector-worker --config
configfile --workers=2
```

The same thing could be done with:

``` bash
$ PYTHONPATH=/path/to/server connector/openerp-connector-worker --config
configfile --workers=all:2
```

`all` is a reserved channel which shouldn't be used in the `job` decorator.

More control is possible using the `--workers` option.

1 worker on `default`, 1 on `system_a`, 2 on `system_b`.

```
--workers=default:1,system_a:1,system_b:2
```

1 worker on `default`, 1 on `system_a.import`, 1 on `system_a.export`, 2 on `system_b`.

```
--workers=default:1,system_a.import:1,system_a.export:1;system_b:2
```

Several channels can be watched by a worker:

So we want 2 workers sharing all the imports of system A and B, and 1 for the export of both systems:

```
[...]

PR #52, opened 2015-03-31
[8.0] job runner

sbidoul:
This is an alternative to connector workers that 
- avoids polling, 
- is easier to deploy, 
- implements job channels (a variation on #43).

For more information on how to use it, see the docstring in runner.py.
For more information on job channels, see the docstring of the Channel class in channels.py.

It is safe to try in real deployments because it is fully compatible and it is easy to switch between the new runner and workers.

TODO:
- [x] can we run the doctests from travis/runbot?
- [x] there is a couple of todo left in the code, but nothing preventing real life usage according to me

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -42,8 +42,8 @@ class RunJobController(http.Controller):
             raise
         return job
 
-    @http.route('/runjob', type='http', auth='none')
-    def handler(self, db, job_uuid, **kw):
+    @http.route('/connector/runjob', type='http', auth='none')
+    def runjob(self, db, job_uuid, **kw):
 
         session_hdl = ConnectorSessionHandler(db,
                                               openerp.SUPERUSER_ID)
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -79,11 +79,9 @@ class RunJobController(http.Controller):
             _logger.debug('%s started', job)
             with session_hdl.session() as session:
                 job.perform(session)
-            _logger.debug('%s done', job)
-
-            with session_hdl.session() as session:
                 job.set_done()
                 self.job_storage_class(session).store(job)
+            _logger.debug('%s done', job)
 
         except NothingToDoJob as err:
             if unicode(err):
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -1,0 +1,116 @@
+import logging
+import traceback
+from cStringIO import StringIO
+
+from psycopg2 import OperationalError
+
+import openerp
+from openerp import http
+from openerp.service.model import PG_CONCURRENCY_ERRORS_TO_RETRY
+
+from ..session import ConnectorSessionHandler
+from ..queue.job import (OpenERPJobStorage,
+                         ENQUEUED)
+from ..exception import (NoSuchJobError,
+                         NotReadableJobError,
+                         RetryableJobError,
+                         FailedJobError,
+                         NothingToDoJob)
+
+_logger = logging.getLogger(__name__)
+
+PG_RETRY = 5  # seconds
+
+
+# TODO: perhaps the notion of ConnectionSession is less important
+#       now that we are running jobs inside a normal Odoo worker
+
+
+class RunJobController(http.Controller):
+
+    job_storage_class = OpenERPJobStorage
+
+    def _load_job(self, session, job_uuid):
+        """ Reload a job from the backend """
+        try:
+            job = self.job_storage_class(session).load(job_uuid)
+        except NoSuchJobError:
+            # just skip it
+            job = None
+        except NotReadableJobError:
+            _logger.exception('Could not read job: %s', job_uuid)
+            raise
+        return job
+
+    @http.route('/runjob', type='http', auth='none')
+    def handler(self, db, job_uuid, **kw):
+
+        session_hdl = ConnectorSessionHandler(db,
+                                              openerp.SUPERUSER_ID)
+
+        def retry_postpone(job, message, seconds=None):
+            with session_hdl.session() as session:
+                job.postpone(result=message, seconds=seconds)
+                job.set_pending(self)
+                self.job_storage_class(session).store(job)
+
+        with session_hdl.session() as session:
+            job = self._load_job(session, job_uuid)
+            if job is None:
+                return ""
+
+        try:
+            # if the job has been manually set to DONE or PENDING
+            # before its execution, stop
+            # TODO: is this still possible?
+            if job.state != ENQUEUED:
+                return
+
+            with session_hdl.session() as session:
+                # TODO: set_started should be done atomically with
+                #       update queue_job set=state=started
+                #       where state=enqueid and id=
+                job.set_started()
+                self.job_storage_class(session).store(job)
+
+            _logger.debug('%s started', job)
+            with session_hdl.session() as session:
+                job.perform(session)
+            _logger.debug('%s done', job)
+
+            with session_hdl.session() as session:
+                job.set_done()
+                self.job_storage_class(session).store(job)
+
+        except NothingToDoJob as err:
+            if unicode(err):
+                msg = unicode(err)
+            else:
+                msg = None
+            job.cancel(msg)
+            with session_hdl.session() as session:
+                self.job_storage_class(session).store(job)
+
+        except RetryableJobError as err:
+            # delay the job later, requeue
+            retry_postpone(job, unicode(err))
+            _logger.debug('%s postponed', job)
+
+        except OperationalError as err:
+            # Automatically retry the typical transaction serialization errors
+            if err.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
+                raise
+            retry_postpone(job, unicode(err), seconds=PG_RETRY)
+            _logger.debug('%s OperationalError, postponed', job)
+
+        except (FailedJobError, Exception):
+            buff = StringIO()
+            traceback.print_exc(file=buff)
+            _logger.error(buff.getvalue())
+
+            job.set_failed(exc_info=buff.getvalue())
+            with session_hdl.session() as session:
+                self.job_storage_class(session).store(job)
+            raise
+
+        return ""
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -60,10 +60,13 @@ class RunJobController(http.Controller):
                 return ""
 
         try:
-            # if the job has been manually set to DONE or PENDING
+            # if the job has been manually set to DONE or PENDING,
+            # or if something tries to run a job that is not enqueued
             # before its execution, stop
-            # TODO: is this still possible?
             if job.state != ENQUEUED:
+                _logger.warning('job %s is in state %s '
+                                'instead of enqueued in /runjob',
+                                job.state, job_uuid)
                 return
 
             with session_hdl.session() as session:
--- a/connector/doc/guides/multiprocessing.rst
+++ b/connector/doc/guides/multiprocessing.rst
@@ -1,9 +1,16 @@
 .. _multiprocessing:
 
 
-######################################
-Use the connector with multiprocessing
-######################################
+##############################################
+Use the connector with multiprocessing workers
+##############################################
+
+.. note:: In a future version, workers will be deprecated
+          in favor of the newer job runner which is more efficient and
+          supports job channels. You should try the job runner first
+          and fall back to using workers in case the runner does not
+          work (sic) for you, in which case we will very much appreciate
+          a github issue describing the problems you encoutered.
 
 When Odoo is launched with 1 process, the jobs worker will run
 threaded in the same process.
--- a/connector/doc/guides/multiprocessing.rst
+++ b/connector/doc/guides/multiprocessing.rst
@@ -10,7 +10,7 @@ Use the connector with multiprocessing workers
           supports job channels. You should try the job runner first
           and fall back to using workers in case the runner does not
           work (sic) for you, in which case we will very much appreciate
-          a github issue describing the problems you encoutered.
+          a github issue describing the problems you encountered.
 
 When Odoo is launched with 1 process, the jobs worker will run
 threaded in the same process.
--- a/connector/jobrunner/__init__.py
+++ b/connector/jobrunner/__init__.py
@@ -1,11 +1,36 @@
+# -*- coding: utf-8 -*-
+##############################################################################
+#
+#     This file is part of connector, an Odoo module.
+#
+#     Author: Stéphane Bidoul <stephane.bidoul@acsone.eu>
+#     Copyright (c) 2015 ACSONE SA/NV (<http://acsone.eu>)
+#
+#     connector is free software: you can redistribute it and/or
+#     modify it under the terms of the GNU Affero General Public License
+#     as published by the Free Software Foundation, either version 3 of
+#     the License, or (at your option) any later version.
+#
+#     connector is distributed in the hope that it will be useful,
+#     but WITHOUT ANY WARRANTY; without even the implied warranty of
+#     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#     GNU Affero General Public License for more details.
+#
+#     You should have received a copy of the
+#     GNU Affero General Public License
+#     along with connector.
+#     If not, see <http://www.gnu.org/licenses/>.
+#
+##############################################################################
+
 import logging
+import os
 from threading import Thread
 import time
 
-from openerp.service import db
 from openerp.service import server
+from openerp.tools import config
 
-from . import channels
 from .runner import OdooConnectorRunner
 
 _logger = logging.getLogger(__name__)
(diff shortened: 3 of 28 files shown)


connector/jobrunner/runner.py
```diff
@@ -0,0 +1,302 @@
...
+                ('channel' if self.has_channel else 'NULL',
+                 'seq' if self.has_seq else 0,
+                 where)
+        with closing(self.conn.cursor()) as cr:
+            cr.execute(query, args)
+            return list(cr.fetchall())
+
+    def set_job_enqueued(self, uuid):
+        with closing(self.conn.cursor()) as cr:
+            cr.execute("UPDATE queue_job SET state=%s, date_enqueued=NOW() "
```
lmignon: the date_enqueued must be truncated to 'seconds'

``` SQL
date_enqueued=date_trunc('seconds', now()::timestamp)
```

otherwise Odoo will not be able to read it and will raise `ValueError`

```
  File "/home/lmi/projects/openerp/openerp-deldrive-buildout/addons-connector/connector/queue/model.py", line 113, in _change_job_state
    job = storage.load(job.uuid)
  File "/home/lmi/projects/openerp/openerp-deldrive-buildout/addons-connector/connector/queue/job.py", line 265, in load
    stored.date_enqueued, DEFAULT_SERVER_DATETIME_FORMAT)
  File "/usr/lib/python2.7/_strptime.py", line 328, in _strptime
    data_string[found.end():])
ValueError: unconverted data remains: .024856
```

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+import logging
+from weakref import WeakValueDictionary
+
+from ..queue.job import PENDING, ENQUEUED, STARTED, FAILED, DONE
+NOT_DONE = (PENDING, ENQUEUED, STARTED, FAILED)
+
+_logger = logging.getLogger(__name__)
+
+
+class PriorityQueue:
```
guewen: Can you change the classes to new-style classes?

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+    >>> j0 == j0
+    True
+    """
+
+    def __init__(self, db_name, channel, uuid,
+                 seq, date_created, priority, eta):
+        self.db_name = db_name
+        self.channel = channel
+        self.uuid = uuid
+        self.seq = seq
```
guewen, line 212: What is seq?

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+            raise IndexError()
+        while True:
+            if not self._heap:
+                raise IndexError()
+            o = self._heap[0]
+            if o in self._removed:
+                o2 = heappop(self._heap)
+                assert o2 == o
+                self._removed.remove(o)
+                self._known.remove(o)
```
guewen, line 88: In that branch is it expected to return `None`?

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+            return self._eta_queue.pop()
+        else:
+            return self._queue.pop()
+
+
+class Channel:
+    """A channel for jobs, with a maximum number of workers.
+
+    Job channels are joined in a hierarchy down to the root channel.
+    When a job channel has free workers, jobs are dequeued, marked
```
guewen: I find the term `workers` confusing here as this word is used as for a different meaning in Odoo (we could argue that the `workers` are how many odoo workers we can speak to at the same time but it would be a bit far-fetched imo)
Would `capacity` be better? Other ideas? And if no better idea I can live with `workers` ;-)

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+    without risking to overflow the system.
+    """
+
+    def __init__(self, name, parent, workers=1, sequential=False):
+        self.name = name
+        self.parent = parent
+        if self.parent:
+            self.parent.children.append(self)
+        self.children = []
+        self.workers = None
```
guewen: `self.workers = workers` I guess

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+        self.sequential = False
```
guewen: `self.sequential = sequential`

Or maybe remove the arguments since they are set afterwards in `configure()`.

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+    are actually sent to Odoo for execution.
+
+    Should a downstream channel have less capacity than its upstream channels,
+    jobs going downstream will be enqueued in the downstream channel,
+    and compete normally according to their properties (priority, etc).
+
+    Using this technique, it is possible to enforce sequence in a channel
+    with 1 worker. It is also possible to dedicate a channel with a
+    limited number of workers for application-autocreated subchannels
+    without risking to overflow the system.
```
guewen, line 331: :+1: great doc

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,602 @@
...
+    def fullname(self):
+        if self.parent:
+            return self.parent.fullname + '.' + self.name
+        else:
+            return self.name
+
+    def get_subchannel_by_name(self, subchannel_name):
+        for child in self.children:
+            if child.name == subchannel_name:
+                return child
```
guewen: Perhaps `self.children` could be a `dict` to avoid this loop.

connector/jobrunner/runner.py
```diff
@@ -0,0 +1,303 @@
...
+                self.channel_manager.remove_db(db_name)
+                db.close()
+            except:
+                _logger.warning('error closing database %s',
+                                db_name, exc_info=True)
+        self.db_by_name = {}
+        for db_name in self.get_db_names():
+            db = Database(db_name)
+            if not db.has_connector:
+                _logger.debug('connector is not installed for db %s', db_name)
```
guewen, line 235: When we install a the addon `connector` on a new db, do we need to restart odoo to have it considered by the runner? (and the reverse when we uninstall it)

connector/controllers/main.py
```diff
@@ -0,0 +1,117 @@
...
+
+        except (FailedJobError, Exception):
+            buff = StringIO()
+            traceback.print_exc(file=buff)
+            _logger.error(buff.getvalue())
+
+            job.set_failed(exc_info=buff.getvalue())
+            with session_hdl.session() as session:
+                self.job_storage_class(session).store(job)
+            raise
```
guewen, line 115: I didn't compared, but from memory the code seems quite similar to the one in https://github.com/OCA/connector/blob/8.0/connector/queue/worker.py#L71. Is it something we can factorize? Maybe it should be moved in the `Job` class?

guewen:
As said before, this is a brilliant work. I added some suggestions or interrogations.
I want to propose a new branch with the channel models / jobs on top of your branch but I fear not to have the time before the 20th of april (holidays from 10th to 19th).
We are testing the 7.0 branch since our customer having the largest number of jobs is on 7.0.

guewen:
> can we run the doctests from travis/runbot?

Here is a PR: https://github.com/acsone/connector/pull/1

Tests will be executed when `--test-enable` is used.

guewen:
BTW could you rebase your branch?

guewen:
Started the implementation of the channels configuration: https://github.com/acsone/connector/pull/2

connector/jobrunner/channels.py
sbidoul: I was thinking the same. Capacity is fine. Adapted.

connector/jobrunner/runner.py
sbidoul, line 235: Currently the server needs to be restarted indeed. I added a TODO about this.

connector/controllers/main.py
sbidoul, line 115: This is a shameless copy/paste indeed. Except I removed some worker-specific logic (related to worker_uuid and eta). This could be factorized and moved to runjob indeed.

connector/jobrunner/channels.py
sbidoul: okay, done.

connector/jobrunner/channels.py
sbidoul, line 331: Thanks! I updated to talk about capacity instead of workers.

connector/jobrunner/channels.py
sbidoul: done

connector/jobrunner/channels.py
sbidoul: done

connector/jobrunner/channels.py
sbidoul, line 212: It's my idea to implement guaranteed sequence on job execution. date_created is not sufficient, and priority is another concept.

It is not fully implemented yet, because jobs that are retried get an eta and this does not work well with sequentiality... needs further thinking.

connector/jobrunner/channels.py
sbidoul, line 88: In that branch we have a removed Item on top of the heap, so we pop it and try again until a non-removed item is found or the heap is empty (in which case we raise IndexError).

sbidoul:
I rebased. Thanks for the review @guewen.

connector/jobrunner/channels.py
guewen, line 88: Yes of course, it is enclosed in a `while True`... Don't know how I could not notice that ;-)

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,629 @@
...
+            else:
+                try:
+                    channel = self.get_channel_by_name(channel_name)
+                except ChannelNotFound:
+                    _logger.warning('unknown channel %s, '
+                                    'using root channel for job %s',
+                                    channel_name, uuid)
+                    channel = self._root_channel
+            job = ChannelJob(db_name, channel, uuid,
+                             seq, date_created, priority, eta)
```
lmignon: Be careful. In Odoo python `date`, `datetime` and `time`  are returned as `string` by psycopg2. (see https://github.com/odoo/odoo/blob/8.0/openerp/sql_db.py#L61 and  http://initd.org/psycopg/docs/extensions.html#psycopg2.extensions.new_type). If you assign a `string` as `eta` to your ChannelJob you'll get a `Comparison error between str and datetime` when popping you job at line #L418

``` python
    while not self.capacity or len(self._running) < self.capacity:
            job = self._queue.pop(now=datetime.now())
```

due to #L275

``` python
    def pop(self, now):
        if len(self._eta_queue) and self._eta_queue[0].eta <= now:
```

sbidoul: @lmignon thanks. I fixed that.

connector/queue/model.py
```diff
@@ -387,3 +403,96 @@ def requeue(self):
...
+
+    @api.model
+    def _setup_complete(self):
+        super(JobFunction, self)._setup_complete()
+        for func in JOB_REGISTRY:
+            if not is_module_installed(self.pool, get_openerp_module(func)):
+                continue
+            func_name = '%s.%s' % (func.__module__, func.__name__)
+            if not self.search_count([('name', '=', func_name)]):
+                self.create({'name': func_name})
```
guewen, line 499: I think to add a new `default_channel` keyword argument in the `@job` decorator, that automatically create a channel for the job if not existing yet and link it with the job. Thus, implementations of connectors would be able to propose default channels (example `root.magento.import` and `root.magento.export` for magento).

lmignon:
:+1:  it's really a great improvement!

connector/queue/model.py
sbidoul, line 499: :+1:

guewen, line 499: @sbidoul here it is https://github.com/acsone/connector/pull/3
I also added tests on the job function and channel models.

guewen:
@sbidoul @lmignon We tested the runner (the 7.0 backport) on a production server with ~15000 jobs per day. We encountered "too many connections" issues with postgresql. I think that's only a matter of misconfiguration on our side. Did you issued similar issues ? Do you know how many connections may the runners use to the fullest?

lmignon:
@guewen  We also encountered the same issue with (+100k jobs/d). If I remember well, you have to change the db_maxconn parameter in your configuration (by default 64).

guewen:
> We also encountered the same issue with (+100k jobs/d). If I remember well, you have to change the db_maxconn parameter in your configuration (by default 64).

We tend to have a number as low as possible in this option in order to avoid to exhaust the max connections of postgres itself. 
Example, we have max_connections=200 in postgresql.conf and 16 workers, we put 12 in `db_maxconn` which is sufficient today (so the max connections we use is 192). If we had to put something > 64, multiplied by the quantity of workers, we would need a huge quantity of postgres connections.

I want to find out how we could quantify the connections necessary for the runner, but I don't know exactly how I should proceed. Any idea?

sbidoul:
From the top of my head, the runner needs exactly one postgres connection per database it is listening to. This comes on top of the connections used by the normal Odoo workers of course.

sbidoul:
rebased

guewen:
We miss one thing: documentation on odoo-connector.com
Perhaps we should add information on http://odoo-connector.com/guides/multiprocessing.html or a dedicated page. And we can add the API docstrings it the doc like that https://raw.githubusercontent.com/OCA/connector/8.0/connector/doc/api/api_queue.rst

I do not consider that as blocking though, as we can do it in a second PR.

@sbidoul Whenever you confirm it is ready to merge, I will do it with a great smile on my face :-)

sbidoul:
@guewen I need to write a documentation page indeed. I suggest you wait for it before merging so I have a higher incentive to squeeze that in my agenda :smile:

guewen:
> I need to write a documentation page indeed. I suggest you wait for it before merging so I have a higher incentive to squeeze that in my agenda :smile:

Okay :laughing:

connector/jobrunner/runner.py
```diff
@@ -0,0 +1,299 @@
...
+                ('channel' if self.has_channel else 'NULL',
+                 where)
+        with closing(self.conn.cursor()) as cr:
+            cr.execute(query, args)
+            return list(cr.fetchall())
+
+    def set_job_enqueued(self, uuid):
+        with closing(self.conn.cursor()) as cr:
+            cr.execute("UPDATE queue_job SET state=%s, "
+                       "date_enqueued=date_trunc('seconds', now()::timestamp) "
```
lmignon: @sbidoul I just noticed that the now() should be at UTC. 

``` sql
date_enqueued = date_trunc('seconds', now()::timestamp at time zone 'utc')
```

The code is therefore sensitive to the timezone defined on the database. To work as expected the timezone on the database should be set to 'UTC'

``` sql
#show timezone;
 TimeZone 
----------
 UTC
(1 row)
```

sbidoul: Good catch. It's fixed now.

connector/jobrunner/__init__.py
```diff
@@ -0,0 +1,98 @@
...
+# Here we monkey patch the Odoo server to start the job runner thread
+# in the main server process (and not in forked workers). This is
+# very easy to deploy as we don't need another startup script.
+# The drawback is that it is not possible to extend the Odoo
+# server command line arguments, so we resort to environment variables
+# to configure the runner (channels mostly).
+
+
+# TODO: this is a temporary flag to enable the connector runner
+enable = os.environ.get('ODOO_CONNECTOR_RUNNER_ENABLE')
```
sbidoul: @guewen @lmignon perhaps we can do the economy of one environment variable by using ODOO_CONNECTOR_CHANNELS here instead. What do you think?

guewen: Yes, sounds good.

sbidoul: I removed the need for ODOO_CONNECTOR_RUNNER_ENABLE, setting ODOO_CONNECTOR_CHANNELS is now sufficient to enable the runner.

sbidoul:
@guewen I just pushed a documentation commit. 

Let me know if it's readable or any other thought about it.

I took the liberty to add Acsone in the list of maintainers next to Akretion as I thought this contrib together with Laurent's previous work justifies it but I don't know the volume of Akretion contribs. If you feel this is not appropriate let me know.

sbidoul:
I rebased to get the latest travis config. 

There are 3 todo's left, which IMO can be left for future work:
- better mechanism to http get asynchronously: normally not a real performance issue, but in case it is perhaps using a threadpool would be a solution
- detect databases on which connector is added/removed: currently the server needs to be restarted in such case, which should not be an issue except for multi-tenant hosters
- finalize sequential channels (needs thinking around ETA), so I did not document that feature

So from my point of view it's ready to merge if and when Travis manages to gives back a green status.

guewen:
> I took the liberty to add Acsone in the list of maintainers next to Akretion as I thought this contrib together with Laurent's previous work justifies it but I don't know the volume of Akretion contribs. If you feel this is not appropriate let me know.

Absolutely deserved.

connector/doc/guides/multiprocessing.rst
```diff
@@ -1,9 +1,16 @@
...
+##############################################
+Use the connector with multiprocessing workers
+##############################################
+
+.. note:: In a future version, workers will be deprecated
+          in favor of the newer job runner which is more efficient and
+          supports job channels. You should try the job runner first
+          and fall back to using workers in case the runner does not
+          work (sic) for you, in which case we will very much appreciate
+          a github issue describing the problems you encoutered.
```
guewen: s/encoutered/encountered/

connector/jobrunner/channels.py
```diff
@@ -0,0 +1,742 @@
...
+    parent channel where they wait for available capacity and so on.
+
+    Job channels can be visualized as water channels with a given flow
+    limit (= capacity). Channels are joined together in a downstream channel
+    and the flow limit of the downstream channel limits upstream channels.::
+
+        ---------------------+
+                             |
+                             |
+         Ch. A W:4,Q:12,R:4  +-----------------------
```
guewen: Could you replace `W` by `C`?

guewen: Could you replace `W` by `C`?

guewen:
I spotted 2 very little things in the doc (inline comments), apart that, all good!

lmignon:
I'm really impressed by this PR and the educational aspects of the doctects associated to each new concept. The connector module is one of my favorite OCA addons and a killer feature for Odoo/OCA. IMO it's ready to be merged. :+1:

guewen:
> I'm really impressed by this PR and the educational aspects of the doctects associated to each new concept. The connector module is one of my favorite OCA addons and a killer feature for Odoo/OCA. IMO it's ready to be merged. :+1:

I can't express it better. Many thanks for this contribution.

sbidoul:
Cool. Thanks guys!

agb80:
@sbidoul  is this runner available on 7.0 branch?

sbidoul:
@agb80 yes, #53 is merged in 7.0

PR #53, opened 2015-04-03
7.0 jobrunner lmi

lmignon:
This is a backport of an alternative to connector workers proposed by Stéphane Bidoul (@sbidoul) on the 8.0 branch (https://github.com/OCA/connector/pull/52)
The main difference is that you need to use the script `connector/openerp-connector-runner` to start the runner. For more information on how to use it, see the docstring into the script.


bealdav:
What a huge work. Nice
Have you considered to not use Odoo name in class, var and constant. Who knows the future ?
If you use ErpMyClass, etc. you can't be wrong for ever.

sbidoul:
@bealdav thanks for the feedback. I removed Odoo from the class name in the 8.0 branch. I prefer to keep it in environment variable, though.

sbidoul:
All, please provide review comments in #52 because it's the main development branch.

guewen:
Thanks for the backport! We are starting to test it since we would be able to try it out in production with a lot of jobs.

connector/jobrunner/runner.py
```diff
@@ -0,0 +1,307 @@
...
+from contextlib import closing
+import logging
+import re
+import select
+import threading
+import time
+
+import psycopg2
+from psycopg2.extensions import ISOLATION_LEVEL_AUTOCOMMIT
+import requests
```
guewen, line 89: `requests` should be added in `external_dependencies`

connector/jobrunner/runner.py
```diff
@@ -0,0 +1,307 @@
...
+* It does not run jobs itself, but asks Odoo to run them through an
+  anonymous /runjob HTTP request [1].
+
+How to use
+----------
+* set the following environment variables:
+  - ODOO_CONNECTOR_RUNNER_ENABLE=1
+  - ODOO_CONNECTOR_CHANNELS=root:4 (or any other channels configuration)
+  - optional if xmlrpc_port is not set: ODOO_CONNECTOR_PORT=8069
+* start Odoo with --load=web,connector and --workers > 1 [2]
```
guewen: Is `--load=web,connector` required in v7?

guewen:
We had a hard-time understanding why `http://localhost:8069/connector/runjob` returned a 404 error.
It is because odoo does only register the controller if the addon contains a `static` directory.
Could you add a `static` directory? (with an empty file named `.gitkeep` since git only indexes files and not directories).

mdietrichc2c:
:+1: after @guewen 's remarks are fixed.

connector/jobrunner/runner.py
sbidoul, line 89: done in 8.0 branch

connector/jobrunner/runner.py
lmignon: Yes it's required to register the controller.

guewen: It is available as soon as there is a `static` folder AFAIK.

sbidoul: I've not checked on 7.0, but on 8.0, I had to add it in --load so the /connector/runjob controller becomes available for anonymous users and independently of any database.

guewen: Ok, maybe it's necessary for this use case.

connector/jobrunner/runner.py
guewen, line 89: > done in 8.0 branch

I reported that on the 7.0 branch because `requests` is already a dependency of Odoo 8.0. 
Anyway, it does not hurt to add in in the 8.0 branch too.

lmignon:
@guewen  @mdietrichc2c I plan to update this PR once the 8.0 will be finalized. I prefer that changes are done on 8.0 and back ported to 7.0. Sorry for the delay..

lmignon:
All commits related to job_runner on 8.0 are now backported to 7.0 except those related to the channel definition in Odoo
- Show the channels full names in the logs https://github.com/OCA/connector/commit/cba40d678eb68ec9271b6324bb18503a3d3061bf
- Add security rights https://github.com/OCA/connector/commit/2cd18e0b1b3902490fd9572a5c62b27e2e1b75bc 
- Allow to configure the channels on the job functions …https://github.com/OCA/connector/commit/b9134e8390da4f3e62c00007cfddc3ff9eb38b7c

guewen:
@lmignon thanks! Isn't something we may want in 7.0 as well?

lmignon:
@guewen for sure. Sometimes I'm a bit lazy and the work is a little more important because the channels are defined with the new API.

lmignon:
@guewen, I'll do the backport of the changes related to the channel definition in odoo tomorrow.

guewen:
> Sometimes I'm a bit lazy and the work is a little more important because the channels are defined with the new API.

Didn't think about that. Valid reason ;-)

lmignon:
@guewen channels are backported to 7.0 I've cherry picked your commits, resolved the conflicts and refactored the code to the old api.

guewen:
Tested, approved :+1:  (but waiting for a green build)

lmignon:
Thank you @guewen

PR #54, opened 2015-04-08
[ADD]2 modules that allow you to manage your dnspod through odoo

liulixia1990:
The DNS Connector module is a base framework that should allow integration with other providers.


guewen:
Hi @liulixia,

This repository is only for the base connector framework. See https://github.com/OCA/connector-interfaces/pull/12#issuecomment-90835373

PR #55, opened 2015-04-13
Calculate eta on job to max(eta)+POSTPONE_INTERVAL

sebastiken:
Made modification to enqueue Jobs calculating max eta. So when jobs are postponed (for example, when _Sales Order_ hasn't been **paid**) they will be enqueued with same _interval_ between them.

Maybe it is not a good implementation, but it was done because when there are too much _Sales Order_ to import, and they are _postponed, __Magento__ receives too much _requests_. With this modification, requests are always performed with **POSTPONED_INTERVAL** between them.

I think this interval should be configured from _User Interface_, but first I need somebody review this **PR** to know if this feature is worth.

Excuse my bad English.

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -51,6 +51,7 @@ STATES = [(PENDING, 'Pending'),
 DEFAULT_PRIORITY = 10  # used by the PriorityQueue to sort the jobs
 DEFAULT_MAX_RETRIES = 5
 RETRY_INTERVAL = 10 * 60  # seconds
+POSTPONE_INTERVAL = 3 * 60  # seconds between postpones
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


guewen:
Hi,

Thanks for your proposal!

Indeed, this is something that I reported too on the connector-magento project so I confirm this is necessary: https://github.com/OCA/connector-magento/issues/26

On the implementation, I will think about it but normally the database related stuff is delegated to the JobStorage class.

sebastiken:
@guewen I think your proposition is correct, at least for my scenario. I will try to work on that if I could next days. If you have something developed about this idea, tell me.  Maybe I can help you.

Thanks for your response.

guewen:
Not exactly the same solution, but https://github.com/OCA/connector/pull/75 is about to be merged and can improve the situation you describe (but it has still to be implemented in the magento connector).

sebastiken:
Ok @guewen that modification of a retry pattern is a really good one.
Thanks.

PR #56, opened 2015-04-16
Don't confuse Nose with a function which is not a test

ccomb:
Hi,

I just added some builders on our buildbot and I got an error with Nose : 
http://buildbot.anybox.fr/builders/connector-postgresql-9.2/builds/1/steps/tests/logs/stdio

```
======================================================================
ERROR: connector.tests.test_related_action.test_unwrap_binding
----------------------------------------------------------------------
  File "/srv/buildslave/buildbot.anybox.fr/buildout-caches/eggs/nose-1.3.6-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: test_unwrap_binding() takes exactly 3 arguments (0 given)

----------------------------------------------------------------------
```

What I understand is that `test_unwrap_binding` is not a real test but a function definition used in the test `test_related_action.test_unwrap_binding`. Am I wrong?

After renaming the function it's ok:
http://buildbot.anybox.fr/builders/connector-postgresql-9.2/builds/3/steps/tests/logs/stdio

### DIFF
--- a/connector/tests/test_related_action.py
+++ b/connector/tests/test_related_action.py
@@ -43,7 +43,7 @@ def task_wikipedia(session, subject):
 
 
 @related_action(action=unwrap_binding)
-def test_unwrap_binding(session, model_name, binding_id):
+def try_unwrap_binding(session, model_name, binding_id):
     pass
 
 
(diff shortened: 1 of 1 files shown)


lmignon:
:+1:

yvaucher:
:+1:

guewen:
Right!
Thanks :+1:

PR #57, opened 2015-04-16
French documentation

ccomb:
Here is an exhaustive french translation of the documentation (even with the API).

We currently have a [sphinx build](http://docs.anybox.fr/connector/fr/) which is generated using [this buildout](https://bitbucket.org/anybox/public_buildbot_buildouts/src/0513d1396ec30f41846f19cfab12d825616b089c/odoo-connector.cfg?at=default) and automatically updated by [these builders](http://buildbot.anybox.fr/waterfall?builder=connector-postgresql-9.2&builder=connector-postgresql-9.3&builder=connector-postgresql-9.4&reload=none) using this [configuration](https://bitbucket.org/anybox/public_buildbot_buildouts/src/c6a1591671770f432e0b7230cc2d780bb14e4bc4/MANIFEST.cfg?at=default#cl-205)


connector/event.py
```diff
@@ -53,7 +53,7 @@ def do_something_product(session, model_name, a, b):
         on_my_event.subscribe(do_something_product,
                               model_names='product.product')
 
-    We can also replace an event::
+    We can also replace an consumer::
```
lmignon: :%s/an consumer/a consumer

lmignon:
:+1:  Thank you @ccomb!!!

ccomb:
I'm adding a short text in the contribute page to explain how to maintain a translation

guewen:
Wow, what a massive work! 
Thanks @ccomb :heart_eyes:

connector/doc/project/contribute.rst
```diff
@@ -59,3 +59,41 @@ New projects should ideally respect it.
...
+- build the translated documentation to html: ``make SPHINXOPTS="-Dlanguage=fr"
+  html``
+
+The same using a `buildout
+<https://bitbucket.org/anybox/public_buildbot_buildouts/src/tip/odoo-connector.cfg>`_::
+
+    $ mkdir buildout && cd buildout
+    $ wget https://bitbucket.org/anybox/public_buildbot_buildouts/raw/tip/odoo-connector.cfg
+    $ wget https://bitbucket.org/anybox/public_buildbot_buildouts/raw/tip/bootstrap.py
+    $ python bootstrap.py && bin/buildout -c odoo-connector.cfg
```
guewen: Following these steps, `python bootstrap.py` gives: `Error: Couldn't open /tmp/tmp/buildout.cfg`. We have to download `odoo-connector.cfg` to `buildout.cfg` before with `wget https://bitbucket.org/anybox/public_buildbot_buildouts/raw/tip/odoo-connector.cfg -O buildout.cfg`.

guewen: Maybe repeat this step here not to miss it:
`Add`this patch https://bitbucket.org/shimizukawa/sphinx-intl/pull-request/9/`_ to sphinx-intl (until merged) to support *fuzzy* translations`

guewen:
@ccomb According to you, what is the better way to publish the french version on odoo-connector.com?
One way I think of is to push it to a `/fr` subfolder and add a link on the index page saying something like "this documentation is also available in french" directing to `/fr`.

gurneyalex:
just wow!

guewen:
Would your builders be able to push the docs on odoo-connector.com (considering it would have access rights) and would you agree to push the builds there? Currently this is a manual process, I `git add`, `git commit` and `git push` the new  builds in the repo: https://github.com/camptocamp/openerp-connector-docs (we can move the repo if required) (then the pages are automatically served by GitHub pages).

ccomb:
@guewen the 'fr/' subfolder is a good idea, since sphinx is still not able to natively host several translations, indeed the index page should provide a list of the available translations pointing to the correct url.
I think our buildbot is capable of uploading the build to a different url, because the upload is just the [step 20](http://buildbot.anybox.fr/builders/connector-postgresql-9.4/builds/7) of the builder, whose configuration is public and the destination URL is [visible](https://bitbucket.org/anybox/public_buildbot_buildouts/src/b1ac741276c21316b20faf1a9372bbcdf605b0c2/MANIFEST.cfg?at=default#cl-16) in the configuration. Just need to cross-check with @gracinet

gracinet:
@guewen previously I had only @comb latest comment on-hand, so I thought it was only a matter of subpath… There's no uploading of any kind currently from our buildmaster to another server (we're cheating, so to speak), but I could add a special step to that effect.

I've never hooked such a bot to GitHub, so if there's a better way than creating a full-blown user account, please tell me

guewen:
> I've never hooked such a bot to GitHub, so if there's a better way than creating a full-blown user account, please tell me

A user account for pushing the changes you mean? I think that's the way to go.

guewen:
@gracinet @ccomb I'm currently investigating the build of the docs from Travis

guewen:
Let get it merged and I will complete #60 to build the french doc as well and include the links.
Thanks again @ccomb

PR #67, opened 2015-05-15
Wrong arguments position in a log message

guewen:


sbidoul:
:+1: 

So you managed to trigger that warning?

guewen:
> So you managed to trigger that warning?

Yes! :-)
I was investigating #68 and called the controller directly on a job such as `/connector/runjob?job_uuid=e3ecfa2e-e425-481c-bc0d-3d1cc64f2c02`

guewen:
BTW would it be "dangerous" if we allowed to call the controller directly on `pending` jobs?

sbidoul:
> BTW would it be "dangerous" if we allowed to call the controller directly
> on pending jobs?
> 
> Someone knowing uuids could then start any job. An external attacker would have to guess uuids to do any harm so maybe it's not dangerous indeed. In that case the worst he could do is a DoS by launching many jobs at once.

Are uuids guessable? If not we could maybe skip the enqueued state.

-sbi

guewen:
I would say very difficult to predict, if possible, but they are not designed _in the aim_ to be unpredictable. Anyway, I was thinking about that when I was debugging, just to check if jobs were runnable, but I see no real application otherwise. So maybe we can discuss about that again if we have a concrete use case for that.

lmignon:
:+1:

pedrobaeza:
:+1:

PR #69, opened 2015-05-16
[IMP] connector-runner: graceful stop mechanism

sbidoul:
The previous version was relying a daemon thread, which was ok, except it would continue to enqueue jobs while the server was stopping, and such jobs would never start.

This in combination with https://github.com/odoo/odoo/pull/6738 should ensure jobs are left in a consistent state upon normal shutdown. Only a hard crash would leave jobs in a started state in which case those must be requeued manually.

### DIFF
--- a/connector/jobrunner/__init__.py
+++ b/connector/jobrunner/__init__.py
@@ -49,49 +49,75 @@ START_DELAY = 5
 enable = os.environ.get('ODOO_CONNECTOR_CHANNELS')
 
 
-def run():
-    # sleep a bit to let the workers start at ease
-    time.sleep(START_DELAY)
-    port = os.environ.get('ODOO_CONNECTOR_PORT') or config['xmlrpc_port']
-    channels = os.environ.get('ODOO_CONNECTOR_CHANNELS')
-    runner = ConnectorRunner(port or 8069, channels or 'root:1')
-    runner.run_forever()
+class ConnectorRunnerThread(Thread):
 
+    def __init__(self):
+        Thread.__init__(self)
+        self.daemon = True
+        port = os.environ.get('ODOO_CONNECTOR_PORT') or config['xmlrpc_port']
+        channels = os.environ.get('ODOO_CONNECTOR_CHANNELS')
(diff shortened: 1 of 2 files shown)


guewen:
Very well observed! I confirm that `started` jobs are no longer interrupted with this patch and odoo/odoo#6738. 

```
2015-05-20 11:44:16,198 14980 DEBUG openerp_dev openerp.addons.connector.controllers.main: <Job 341d9f10-09b4-40c7-9565-c17a09a8d1c0, priority:10> done
2015-05-20 11:44:16,202 14980 INFO openerp_dev openerp.service.server: Worker (14980) exiting. request_count: 8, registry count: 1.
```

Great! Just waiting on an answer on odoo/odoo#6738 now...

lmignon:
:+1:

guewen:
@lmignon did you backported  this PR in v7 too?

lmignon:
@guewen not yet. it will be done tomorrow as I need it

sbidoul:
In v7 the runner is a separate process? In which case it should not be necessary to backport this: killing the runner before OpenERP should be sufficient.

guewen:
> In v7 the runner is a separate process? In which case it should not be necessary to backport this: killing the runner before OpenERP should be sufficient.

It makes sense.

@sbidoul If I understand correctly, we can already merge this one even if odoo/odoo#6738 is not merged yet. It won't prevent jobs to be interrupted, but it will prevent to push new jobs when exiting, is it correct?

lmignon:
@sbidoul you're right. The one I need to back port is https://github.com/odoo/odoo/pull/6738

sbidoul:
> @sbidoul https://github.com/sbidoul If I understand correctly, we can
> already merge this one even if odoo/odoo#6738
> https://github.com/odoo/odoo/pull/6738 is not merged yet. It won't
> prevent jobs to be interrupted, but it will prevent to push new jobs when
> exiting, is it correct?
> 
> Correct.

guewen:
:+1: 
Thanks!

PR #74, opened 2015-06-01
Deprecate install_in_connector()

jssuzanne:
See if we can remove install_in_connector() by checking in ir.module.module with caching of the result


connector/module.py
```diff
@@ -0,0 +1,39 @@
...
+    @ormcache(skiparg=1)
+    def is_module_installed(self, module_name):
+        return bool(len(self.env['ir.module.module'].search([
+            ('name', '=', module_name),
+            ('state', '=', 'installed')])))
+
+    def state_update(self, *args, **kwargs):
+        res = super(IrModuleModule, self).state_update(*args, **kwargs)
+        self.clear_caches()
+        return res
```
StefanRijnhart: In module_uninstall(), in addons/base/module/module.py the state is set to 'uninstalled' without calling this method. Do you want to cover this scenario? I'm guessing the previous implementation didn't so it would not be a regression.

jssuzanne: You are right, I will fix this now, thank

StefanRijnhart:
CLA false alarm. Github handle is registered in the OCA instance and CLA's signed. Maybe someone added the Github handle to the OCA partner record just now?

connector/producer.py
```diff
@@ -45,7 +46,7 @@
 @openerp.api.returns('self', lambda value: value.id)
 def create(self, vals):
     record_id = create_original(self, vals)
-    if self.pool.get('connector.installed') is not None:
+    if is_module_installed(self.env, 'connector') is not None:
```
StefanRijnhart: 'if not is_module_installed()' is preferred.

Curious to see you passing self.env instead of self. Am I missing something?

StefanRijnhart: Oh, probably something to do with this being a static method you are importing.

jssuzanne: No, I just replace pool by env in the goal to get the cursor, If you prefer I can replace self.env by self

StefanRijnhart: If that works, it would be clearer. You are passing self.env as the first argument of  https://github.com/OCA/connector/pull/74/files#diff-2700ffb2a55d3a936589a6fab4428330R31 called self, then access self.env. Again, I am wondering why this works at all.

jssuzanne: You, not see the good file.
At the top of the file you can read the import::

``` python
     from .connector import is_module_installed
```

this function is an entry point, and this entry point is tested.

guewen:
Great! Thanks
:+1:

connector/module.py
```diff
@@ -0,0 +1,54 @@
...
+    The next methods change the state of the module, then they must invalidate
+    the cache of 'is_module_installed' method:
+
+    * state_update
+    * module_uninstall
+
+    """
+    _inherit = 'ir.module.module'
+
+    @ormcache(skiparg=1)
```
guewen: Wondering if it shouldn't be:

```
@api.model
@ormcache()
def is_module_installed(self, module_name):
    ....
```

Because the default decorator is `@api.multi`, that's why it you need to set `skiparg=1` where the arg is `ids`, always empty in this case.

jssuzanne: Unfortunately, your solution is not working, because if you add both decorator, the method doesn't appear in the registry

guewen: `@api.model` would be more correct but does not work with `@ormcache`. So okay to keep it like that.

StefanRijnhart:
I have now tested it, and I am afraid this is not working as expected. After installation, the cache still returns False. Only after a restart of Odoo, the method returns True. After deinstallation, any create results in an Exception 'ir.module.module' object has no attribute 'is_module_installed'. The same is true when I try to use other databases on the instance that don't have the module installed, so these are in fact unusable.

Maybe it is better to test for the presence of an attribute on one of the modules that the connector module adds, like I suggested before?

connector/connector.py
```diff
@@ -49,42 +46,18 @@ def _get_openerp_module_name(module_path):
...
+    log_deprecate("This call to 'install_in_connector()' has no effect and is "
+                  "not required.")
 
-    It has to be called once per OpenERP module to plug.
 
-    Under the cover, it creates a ``openerp.models.AbstractModel`` whose
-    name is the name of the module with a ``.intalled`` suffix:
-    ``{name_of_the_openerp_module_to_install}.installed``.
+def is_module_installed(env, module_name):
+    if env.registry.get('connector.backend') is None:
```
StefanRijnhart: Do you still need the other test? What happens if you just return (a boolean cast of) the value of env.registry.get('connector.backend')?

StefanRijnhart:
Much better now, thanks! No more errors after deinstallation or on other databases. Only the installation of the module is still not registered properly. Odoo needs a restart to clear the cache.

connector/connector.py
jssuzanne: the verification of 'connector.backend' is use only to test if connector is installed:
if not return False
if True, call the [ir.module.module].is_module_installed method, come from connector, to check if the `module_name` is installed

StefanRijnhart: So, the generic use case does not work otherwise we would not need to check the existence of the model. And the second test is made redundant by the first test for this specific use case.

jssuzanne: The first test is only to check that the connector module is installed, but we want make this test with another model too

bguillot:
looks good, thanks 
:+1:

StefanRijnhart:
Works perfect now, thanks! :+1:

guewen:
@jssuzanne a test is failing. I corrected it on https://github.com/jssuzanne/connector/pull/1, could you merge it?

guewen:
Green!
Thanks!

PR #81, opened 2015-06-16
Activate transifex

guewen:

PR #82, opened 2015-06-16
Correct PR #81

guewen:
DOCS is set to 0 or 1 so check that and not if DOCS is set/unset

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,9 +36,9 @@ services:
   - postgresql
 
 script:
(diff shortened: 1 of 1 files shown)


pedrobaeza:
These strange shell commands... :+1:

guewen:
Transifex says 
> WARNING! Transifex password not recognized- exiting early.

yvaucher:
:+1:

lmignon:
:+1:

guewen:
This time it is green

PR #84, opened 2015-06-26
Option --load=web,connector no longer required

guewen:
If the module contains a folder named 'static', the controller will be loaded.
So it is no longer necessary to load it with the --load option

### DIFF
--- a/connector/jobrunner/__init__.py
+++ b/connector/jobrunner/__init__.py
@@ -25,7 +25,6 @@
 
 import logging
 import os
-from threading import Thread
 import time
 
 from openerp.tools import config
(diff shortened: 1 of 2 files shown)


guewen:
@lmignon I need your review :-)

agb80:
Does this PR means that there is no longer requirement for two Odoo running process for get runner working on v7.0?

lmignon:
:+1:

sbidoul:
:+1: nice trick.

@guewen will that work on 8.0 too?

@agb80 in 7.0 the runner remains in a separate process. In 8.0 it is a thread in the main odoo process.

guewen:
@sbidoul I guess `--load` is still required in v8 to be able to be able to patch the server (https://github.com/OCA/connector/blob/8.0/connector/jobrunner/__init__.py#L72-L75) but I didn't tested so it would worth to test.

agb80:
@sbidoul let me understand this, whit out this patch on v7.0 for make the connector work with runner you need: launch the server with --load=web,connector && launch the runner from its own script?

If running server with out load option could explain that my jobs stuck on "pending" or "running" state?

guewen:
@agb80 exactly

laetitia-gangloff:
:+1:

PR #86, opened 2015-07-02
7.0 - manage retryable job with runner

laetitia-gangloff:
With runner,

When a retryablejoberror is raised, the retry job is not incremented.
It is reset to zero when set_pending is called, in consequence job retry indefinitely.

Any advices or examples to test controller is welcome !


lmignon:
:+1: (core review + functional test)

guewen:
Thanks! :+1:
I fast track

PR #87, opened 2015-07-02
8.0 - manage retryable job with runner

laetitia-gangloff:
forward port of #86


lmignon:
:+1:  Without this change, RetryAble jobs will be retried indefinitely.

guewen:
:+1: same than in 7.0, fast track

PR #89, opened 2015-07-02
Add an index on queue_job.worker_id

guewen:
This column appears often in the where clauses with the old workers.


pedrobaeza:
Also :+1:

lmignon:
Index thumb up day :+1:

laetitia-gangloff:
:+1:

PR #76, opened 2015-06-02
Default binder

ccomb:
Default Binder implementation to avoid rewriting it in each connector

PR #91, opened 2015-07-03
Default binder

guewen:
Follows #76.


guewen:
This is the change I want to promote: https://github.com/OCA/connector/commit/1525a4ad06c0d02f78dbdc1568a7ab4274a03e32

As a call to `browse()` does not hit the database until we first access one of the fields (other than `id`), I think it is better to always returns a recordset. The API will be easier to use and it reduces the chances of developer errors, we have only one possible type in the return and one knows that she has always to call `.id` on the result of `to_openerp()` when she wants the `id` of a binding. It is also inline with the new API which returns recordsets instead of ids in `search()`.

lmignon:
:+1: (code review)

guewen:
Rebased. I would :heart:  to get this PR merged.

sebastienbeau:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: :+1: 
Yeeesssss, I totaly approve this PR

bguillot:
of course :+1: !
I missed this one I approved the previous one ...

PR #92, opened 2015-07-03
Add retry pattern for jobs

guewen:
Follows #75


guewen:
This change has been accepted here: #75 
I just removed a field which was unused.
I'll merge as soon as I get a green.

PR #94, opened 2015-07-14
Mapper: add modifier to follow m2o relations

LeartS:
A modifier for direct mappings to follow `many2one` fields, a use case I think it's quite common.

Example with model = `product.product`

```
direct = [
    (follow_m2o_relations('product_tmpl_id.categ_id.name'), 'category_name')]
```

### DIFF
--- a/connector/tests/test_mapper.py
+++ b/connector/tests/test_mapper.py
@@ -13,6 +13,7 @@ from openerp.addons.connector.unit.mapper import (
     changed_by,
     only_create,
     convert,
+    follow_m2o_relations,
     m2o_to_backend,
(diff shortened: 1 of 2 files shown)


guewen:
Thanks :-)
A good idea!

Travis complains about:

```
./connector/unit/mapper.py:251:80: E501 line too long (80 > 79 characters)
```

Would you mind to add a test? You can use this test as a basis: https://github.com/OCA/connector/blob/8.0/connector/tests/test_mapper.py#L389-L400

guewen:
Thanks ! :smile: 
:+1:

gurneyalex:
:+1:

Issue #102, opened 2015-07-24
[8.0] RETRY_INTERVAL error (unsupported type for timedelta seconds component: NoneType)


lefeverd:
Hello,
I encountered a bug when a RetryableJobError is raised.
The retry_postpone method is then called, without the keyword arg seconds.
This raises an error in job.py, method postpone:
TypeError: unsupported type for timedelta seconds component: NoneType

**Steps to reproduce**
- Create a job
- Raise a RetryableJobError

**Current Behaviour**
Traceback because of timedelta not allowing NoneType

**Expected Behaviour**
No traceback, job postponed.

guewen:
Closed by #103

PR #103, opened 2015-07-24
[FIX] seconds defaults to RETRY_INTERVAL when no retry_pattern

lefeverd:
Merge Proposal fixing issue #102

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -582,6 +582,8 @@ class Job(object):
                     seconds = postpone_seconds
                 else:
                     break
+        elif not seconds:
+            seconds = RETRY_INTERVAL
         return seconds
 
     def postpone(self, result=None, seconds=None):


guewen:
Perfect! Thanks :+1:

lmignon:
:+1:  seems this error only occurs with the old worker.

lefeverd:
@oca-clabot CLA is on its way.
@lmignon I had the error also with the jobrunner (if that's what you meant).
@guewen Thanks to you!
On an unrelated subject I had an issue when trying to set a retry_pattern in the decorator of a job like:
job(retry_pattern={0: 20})
It is not taken into account, and even when I break in the job decorator, the parameter is None... Maybe I missed something? Should I open an issue?

guewen:
@lefeverd ouch you are right

Here: https://github.com/OCA/connector/blob/8.0/connector/queue/job.py#L702
The argument should be passed along...

Would be marvellous if a few tests for that could be added :relaxed:

sbidoul:
ICLA ok

lefeverd:
@guewen Indeed, I missed it, thanks.
Do you already plan to fix it? If not I can propose a PR (and maybe include some tests).
@sbidoul Thank you!

guewen:
@lefeverd I'm leaving this evening for 2 weeks with no or nearly no internet access so I would be glad if you could propose a PR

Issue #107, opened 2015-08-19
Feature: Using custom ConnectorEnvironment


maljac:
**Scenario**
When using the connector framework you have to use some API wrapper for your external system. This API wrapper maybe stateless or not. If your API wrapper is statefull, how do you pass the api around without creating the API instance and connecting each time again (which is costly)?

E.g. for reading the record, for reading children, for reading translation etc....

**Workaround**

In order to reuse the API instance, I wrote a custom ConnectorEnvironment which gets the API instance at instantiation:

```
class APIConnectorEnvironment(ConnectorEnvironment):
    def __init__(self, backend_record, session, model_name, api=None):
        super(APIConnectorEnvironment, self).__init__(
            backend_record, session, model_name)
        self.api = api

def get_environment(...):
    ...
    env = APIConnectorEnvironment(...., api=api)
    ....
    return env
```

Unfortunately this does not work all the time. In some situations the framework creates a new ConnectorEnvironment and it naturally does not use my custom ConnectorEnvironment class because of the `unit_for` method in the `ConnectorUnit`:

```
def unit_for(self, connector_unit_class, model=None):
        ...
        if model is None or model == self.model._name:
            env = self.connector_env
        else:
            env = ConnectorEnvironment(self.backend_record,
                                       self.session,
                                       model)
        return env.get_connector_unit(connector_unit_class)
```

So, in order to get my custom ConnectorEnvironment working I have to 

1) create a Mixin which must be used for all classes that inherit from `ConnectorUnit` (Exporter, Importer, Deleter, Synchronizer, Mapper, ...)
2) overwrite the `unit_for` method and _replace_ it in the mixin

**Feature Request**

For case (2) it would be nicer to have a hook method in `ConnectorUnit` like:

```
def unit_for(...):
        if model is None or model == self.model._name:
            env = self.connector_env
        else:
            env = self._create_connector_environment(model)
        return env.get_connector_unit(connector_unit_class)

    def _create_connector_environment(self, model):
        return ConnectorEnvironment(self.backend_record, self.session, model)
```

What solutions exist for (1) except for monkey patching ? Did I overlook something or should this be a supported feature: Interchangeable ConnectorEnvironment classes?

guewen:
That's something I already thought about, but never had the use case so never took time to work on that. Definitely :+1: for implementing that.

I think that the hook in `ConnectorUnit` would not be ideal because we would still need to override it.

What about:
1. Create a new instance from the same class than the current env
[...]

PR #108, opened 2015-08-20
Feature: Using custom ConnectorEnvironment

maljac:
Related to #107

### DIFF
--- a/connector/CHANGES.rst
+++ b/connector/CHANGES.rst
@@ -6,7 +6,7 @@ Future
 
 * method 'install_in_connector' is now deprecated
 * Add a retry pattern for jobs (https://github.com/OCA/connector/pull/75)
-* Use custom connector environments and instatiate them with needed attributes (https://github.com/OCA/connector/pull/108)
+* Use custom connector environments and instantiate them with needed attributes (https://github.com/OCA/connector/pull/108)
 
 3.1.0 (2015-05-15)
 ~~~~~~~~~~~~~~~~~~
--- a/connector/CHANGES.rst
+++ b/connector/CHANGES.rst
@@ -6,6 +6,7 @@ Future
 
 * method 'install_in_connector' is now deprecated
 * Add a retry pattern for jobs (https://github.com/OCA/connector/pull/75)
+* Use custom connector environments and instatiate them with needed attributes (https://github.com/OCA/connector/pull/108)
 
 3.1.0 (2015-05-15)
 ~~~~~~~~~~~~~~~~~~
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -268,7 +268,7 @@ class ConnectorEnvironment(object):
     .. attribute:: _propagate_kwargs
 
         List of attributes that must be used by
-        :py:class:`connector.connector.ConnectorEnvironment.create_environment`
+        :py:meth:`connector.connector.ConnectorEnvironment.create_environment`
         when a new connector environment is instantiated.
     """
 
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -342,7 +342,6 @@ class ConnectorEnvironment(object):
         :type connector_env:
             :py:class:`connector.connector.ConnectorEnvironment`
         """
-        # cls = self.__class__
         if connector_env:
             kwargs = {key: getattr(connector_env, key)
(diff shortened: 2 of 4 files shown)


lmignon:
@maljac Thank you for the contribution. Can you provide some unit tests to stress the new functionality and fix the pep8 issue?

```
======== Testing test_flake8 ========
./connector/connector.py:217:21: E126 continuation line over-indented for hanging indent
```

connector/connector.py
```diff
@@ -316,6 +326,29 @@ def get_connector_unit(self, base_class):
         return self.backend.get_class(base_class, self.session,
                                       self.model_name)(self)
 
+    @classmethod
+    def create_environment(cls, backend_record, session, model,
+                           connector_env=None):
```
guewen, line 332: When a `connector_env` is given, we already know the `backend_record`, the `session` and the `model` (which may change) so we can wonder if they could not all be optional (but mandatory when we have no `connector_env`). Still, thinking about it, it would make the body of the class method and the understanding of the API more complicated so it maybe doesn't worth the gain of removing the repetition.

maljac, line 332: Yeah, I was thinking about the same. But "explicit is better than implicit", so as you said, the API would be more complicated.

connector/connector.py
```diff
@@ -261,8 +264,16 @@ class ConnectorEnvironment(object):
     .. attribute:: model_name
 
         Name of the OpenERP model to work with.
+
+    .. attribute:: _propagate_kwargs
+
+        List of attributes that must be used by
+        :py:class:`connector.connector.ConnectorEnvironment.create_environment`
+        when a new connector environment is instiated.
```
guewen: small typo s/instiated/instantiated/

connector/connector.py
```diff
@@ -316,6 +327,29 @@ def get_connector_unit(self, base_class):
...
+                           connector_env=None):
+        """ Create a new environment ConnectorEnvironment.
+
+        :param backend_record: browse record of the backend
+        :type backend_record: :py:class:`openerp.models.Model`
+        :param session: current session (cr, uid, context)
+        :type session: :py:class:`connector.session.ConnectorSession`
+        :param model_name: name of the model
+        :type model_name: str
+        :param connector_env:
```
guewen: Suggestion: _an existing environment from which the kwargs will be propagated to the new one_
Feel free to explain it in another way or improve my suggestion

guewen:
Thanks for this excellent and well realized contribution!
Before a merge, could you add yourself in the authors file (https://github.com/OCA/connector/blob/8.0/connector/AUTHORS) and add a line describing your change in the changelog (https://github.com/OCA/connector/blob/8.0/connector/CHANGES.rst in the "Future" section)?

lmignon:
Thank you @maljac  :+1:

guewen:
I propose you to amend your last commit and to remove the `(tiny change)` on your name ;-)
Thanks :+1:

maljac:
Since you are the project lead, I leave the decission up to you to decide what is a tiny change or not ;-)
commit amended

guewen:
The one who makes a complete PR with docstring and tests can't be granted a tiny change ;-)

We just have to wait some days or have a third approval before we can merge according to the OCA's rules.

connector/connector.py
```diff
@@ -261,8 +264,16 @@ class ConnectorEnvironment(object):
     .. attribute:: model_name
 
         Name of the OpenERP model to work with.
+
+    .. attribute:: _propagate_kwargs
+
+        List of attributes that must be used by
+        :py:class:`connector.connector.ConnectorEnvironment.create_environment`
```
guewen: this one should be `:py:meth:`

connector/CHANGES.rst
```diff
@@ -6,6 +6,7 @@ Future
 
 * method 'install_in_connector' is now deprecated
 * Add a retry pattern for jobs (https://github.com/OCA/connector/pull/75)
+* Use custom connector environments and instatiate them with needed attributes (https://github.com/OCA/connector/pull/108)
```
guewen: s/instatiate/instanciate/

mdietrichc2c:
:+1: on this great PR

guewen:
Thanks for your work!

PR #118, opened 2015-09-15
connector : dynamic channel by company

laetitia-gangloff:
On channel add "channel_by_company" flag to execute job in different channel in function of the company of the job

Use case :
Delivery Order (DO) are transfered by a job.
If many DO contains same product, the transfer failed.
For the moment, these job are retryable, but in some case (many DO) it is not sufficient.
So, we want use a channel with a capacity of 1.
But it is multi-company environment, so we want parallelize transfer of different companies.

### DIFF
--- a/connector/queue/model.py
+++ b/connector/queue/model.py
@@ -29,6 +29,7 @@ from openerp.osv import orm, fields
 from openerp.tools import DEFAULT_SERVER_DATETIME_FORMAT
 from openerp.tools.translate import _
 from openerp import exceptions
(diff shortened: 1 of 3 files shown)


sbidoul:
Hm. I understand the use case and the implementation looks fine.

However I wonder if it's not a bit too specific to go as is in the connector.

What would be of generic interest, though is a mecanism to let the developer select the channel at runtime when creating the job.

lmignon:
The generic approach is implemented by https://github.com/OCA/connector/pull/257

PR #124, opened 2015-10-06
Allow to define seconds and ignore retries when raising RetryableJobError

guewen:
The job will be retried after the given number of seconds if the exception is
raised with a defined 'seconds' argument.


lmignon:
Good Idea! The tests need to be fixed.

lmignon:
:+1:  (Code review) Thank you @guewen

gurneyalex:
:+1:

sebalix:
Great, I need this feature currently, :+1:

PR #130, opened 2015-10-15
Check if a module is installed from the registry

### LINKED ISSUE #85: is_module_installed() alternative
Hi, @guewen 

Since the deprecated install_in_connector(), I've tried to search a replacement like here.

https://github.com/akretion/base-import-mapping/pull/1/files#diff-823631ecb7bd71a0feea07153d79aa38R196

I haven't saw your new method is_module_installed, oups, but 

``` python
if module_name in list(self.env.registry._init_modules):
```

works like a charm.

It seems to be a low level attribute defined here
https://github.com/odoo/odoo/blob/8.0/openerp/modules/registry.py#L56

Don't you think it could be used safely as an alternative of is_module_installed() at least for future Connector version ?

cc @ccomb @StefanRijnhart @sebastienbeau @rvalyi

guewen:
Instead of `ir.module.module`. Simplifies as we never hit the database nor have
to maintain a cache.

Also, it fixes an issue when we run tests directly at the installation of a
module, during the tests the previous implementation was considering the module
as uninstalled instead of installed.

Change proposed by @bealdav. Thanks!

Fixes #85


pedrobaeza:
:+1:

gurneyalex:
nice :+1:

sebalix:
:+1:

bealdav:
thanks @guewen to do this PR instead of me. I like when you remove code. 
It seems to me when I read, that I'm better than I'm in reality ;-)
1 line remove 30 lines, nice
:+1:

sebastienbeau:
:+1:

guewen:
> thanks @guewen to do this PR instead of me. I like when you remove code.
> It seems to me when I read, that I'm better than I'm in reality ;-)
> 1 line remove 30 lines, nice

It is all yours yet!

PR #134, opened 2015-10-30
Update mailing list address

guewen:


lmignon:
:+1:

README.md
```diff
@@ -8,8 +8,8 @@ Odoo Connector Modules
 
 This OpenERP add-on has a modular and generic core, with the ability to be extended with additional modules for new features or customizations.
 
-The team for the community is here: https://launchpad.net/~openerp-connector-community
-Please join and subscribe to the mailing list.
+Feel free to subscribe on the mailing list:
```
pedrobaeza: You should specify in which list inside the page (explicit better than implicit).

lmignon: IMO it's a way to also promote others ML...

guewen: Good point @lmignon 
I changed my pull requests now though :-(

pedrobaeza: Well, all lists are available in that link, but we should guide people to which ML is the adequate for this, because you know that sometimes users can be very "confused"

lmignon: @pedrobaeza it's ok for me.

PR #137, opened 2015-11-02
[8.0][FIX] when a job fails, inactive users are added to its followers

Olivier-LAURENT:


guewen:
Fine!

Thanks for the fix and for including a test! :+1:

pedrobaeza:
:+1:

lmignon:
:+1:

PR #139, opened 2015-11-03
Rename try_advisory_lock to advisory_lock_or_retry

guewen:
It is no longer a context manager, because we would expect the lock to be
released at the end of the 'with' statement but it lasts until the end of the
transaction.

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -231,29 +231,37 @@ class ConnectorUnit(object):
         log_deprecate('renamed to binder_for()')
         return self.binder_for(model=model)
 
-    @contextmanager
-    def try_advisory_lock(self, lock, retry_seconds=1):
-        """ Context manager, tries to acquire a Postgres transactional
-        advisory lock.
+    def advisory_lock_or_retry(self, lock, retry_seconds=1):
+        """ Acquire a Postgres transactional advisory lock or retry job
 
-        If the lock cannot be acquired, it raises a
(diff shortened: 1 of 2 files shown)


guewen:
As discussed in #139 /cc @lmignon @pedrobaeza @damdam-s 
I renamed the method in `advisory_lock_or_retry` which reflects better the behavior I think.

pedrobaeza:
You should change tests accordingly the rename:

``` shell
2015-11-03 07:42:01,583 22882 ERROR openerp_test openerp.addons.connector.tests.test_connector:   File "/home/travis/build/OCA/connector/connector/tests/test_connector.py", line 229, in test_concurrent_import_lock
2015-11-03 07:42:01,583 22882 ERROR openerp_test openerp.addons.connector.tests.test_connector:      with connector_unit.try_advisory_lock(lock):
2015-11-03 07:42:01,583 22882 ERROR openerp_test openerp.addons.connector.tests.test_connector: ` AttributeError: 'ConnectorUnit' object has no attribute 'try_advisory_lock'
```

guewen:
> You should change tests accordingly the rename:

This is what happen when I code before being totally awake, and you caught it even before I could realize :-D
Tests are corrected.

lmignon:
:+1:  Thank you for the changes.

pedrobaeza:
Thanks :+1:

damdam-s:
:+1:

Issue #148, opened 2015-11-18
Job stuck in 'enqueued'


guewen:
We had this case today with the jobrunner (version 7.0).

A job was stuck in `enqueued`, blocking the queue because it took a place, considered as a running job.

Our analysis was that the jobrunner updated the state of the job to enqueued, but the `GET` call to run the job on the Odoo workers had never reached Odoo (no trace of `GET` for this job in the logs). The reason why the `GET` call was not done is unknown (other jobs were correctly processed just before).

In the runner's logs, we have:

```

2015-11-17 18:25:27,873 9168 INFO ? openerp.addons.connector.jobrunner.runner: asking Odoo to run job a9e5d41a-433f-43e5-b584-e6991cf2a64e on db openerp_prod_xxx
2015-11-17 18:25:27,876 9168 INFO ? requests.packages.urllib3.connectionpool: Starting new HTTP connection (1): localhost
```

So we expect to have this in the Odoo logs:

```
2015-11-17 18:25:27,876 5327 INFO openerp_prod_xxx werkzeug: 127.0.0.1 - - [17/Nov/2015 18:25:27] "GET /connector/runjob?db=openerp_prod_xxx&job_uuid=a9e5d41a-433f-43e5-b584-e6991cf2a64e HTTP/1.1" 200 -
```

But there is nothing.

In a general manner, if anything goes wrong when we call `_async_http_get` (https://github.com/OCA/connector/blob/9.0/connector/jobrunner/runner.py#L134-L149) (which is very rare but could happen), a job will stay stuck in `enqueued` state.

A possible workaround would be to add a cron that check if a job is enqueued for too long and requeue it. But I wonder if we couldn't find a nicer way to fix it (what about removing the `enqueued` state? side-effects?)

@sbidoul @lmignon  you surely have ideas :-)

lmignon:
@guewen In the `_async_http_get` exception are ignored . Am I wrong if I say that if an exception occur, the job will stay in the  'enqueued' state and block the queue?   IMO in the `_async_http_get` method we should set the job state to 'failed' if an error occurs and the state is 'enqueued'.

guewen:
> @guewen In the _async_http_get exception are ignored . Am I wrong if I say that if an exception occur, the job will stay in the 'enqueued' state and block the queue? IMO in the _async_http_get method we should set the job state to 'failed' if an error occurs and the state is 'enqueued'.

[...]

PR #150, opened 2015-11-18
Set job back to 'pending' in case of exception

mdietrichc2c:
This PR is for issue #148 in 7.0 ; the goal is to set back the job in state `pending` if execution of the asynchronous job raised an exception (timeout or other).

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -90,7 +90,7 @@ import requests
 
 import openerp
 
-from .channels import ChannelManager, ENQUEUED, NOT_DONE
+from .channels import ChannelManager, PENDING, ENQUEUED, NOT_DONE
 
 SELECT_TIMEOUT = 60
 ERROR_RECOVERY_DELAY = 5
@@ -98,19 +98,34 @@ ERROR_RECOVERY_DELAY = 5
 _logger = logging.getLogger(__name__)
 
 
-def _async_http_get(url):
+def _async_http_get(port, db_name, job_uuid):
+    # Method to set failed job (due to timeout, etc) as pending,
+    # to avoid keeping it as enqueued.
+    def set_job_pending():
+        conn = psycopg2.connect(openerp.sql_db.dsn(db_name))
+        conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+        with closing(conn.cursor()) as cr:
+            cr.execute(
+                "UPDATE queue_job SET state=%s, "
+                "date_enqueued=NULL, date_started=NULL "
+                "WHERE uuid=%s and state=%s", (PENDING, job_uuid, ENQUEUED)
+            )
+
     # TODO: better way to HTTP GET asynchronously (grequest, ...)?
     #       if this was python3 I would be doing this with
     #       asyncio, aiohttp and aiopg
     def urlopen():
+        url = ('http://localhost:%s/connector/runjob?db=%s&job_uuid=%s' %
+               (port, db_name, job_uuid))
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
             requests.get(url, timeout=1)
         except requests.Timeout:
-            pass
+            set_job_pending()
         except:
             _logger.exception("exception in GET %s", url)
+            set_job_pending()
     thread = threading.Thread(target=urlopen)
     thread.daemon = True
     thread.start()
@@ -249,9 +264,7 @@ class ConnectorRunner(object):
             _logger.info("asking Odoo to run job %s on db %s",
                          job.uuid, job.db_name)
             self.db_by_name[job.db_name].set_job_enqueued(job.uuid)
-            _async_http_get('http://localhost:%s'
-                            '/connector/runjob?db=%s&job_uuid=%s' %
-                            (self.port, job.db_name, job.uuid))
+            _async_http_get(self.port, job.db_name, job.uuid)
 
     def process_notifications(self):
         for db in self.db_by_name.values():


connector/jobrunner/runner.py
```diff
@@ -98,19 +98,33 @@
...
+def _async_http_get(port, db_name, job_uuid):
+    # Method to set failed job (due to timeout, etc) as pending,
+    # to avoid keeping it as enqueued.
+    def set_job_pending():
+        conn = psycopg2.connect(openerp.sql_db.dsn(db_name))
+        conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
+        with closing(conn.cursor()) as cr:
+            cr.execute("UPDATE queue_job SET state=%s, "
+                       "date_enqueued=NULL, date_started=NULL, worker_uuid=NULL "
+                       "WHERE uuid=%s", (PENDING, job_uuid))
```
lmignon: we should add 'state='enqueued'` to the where clause

guewen: Yes, because if the controller changes the state to `started` and we get a timeout or no response, the jobrunner could reset a running job to `enqueued` (not sure that could really happen but anyway that's safer).

connector/jobrunner/runner.py
```diff
@@ -98,19 +98,33 @@
...
     def urlopen():
+        url = ('http://localhost:%s/connector/runjob?db=%s&job_uuid=%s' %
+               (port, db_name, job_uuid))
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
             requests.get(url, timeout=1)
         except requests.Timeout:
-            pass
+            _logger.exception("timeout in GET %s", url)
```
guewen: Could you change this log level to `error`? We don't need the traceback on this one.

mdietrichc2c:
@lmignon @guewen thanks for the feedback ; I did the requested changes.

guewen:
Thanks for the fix! :+1:

lmignon:
@mdietrichc2c Thank you for the changes! The code looks good to me. :+1:

laetitia-gangloff:
It seems ok :+1: 
Thank you !

guewen:
Not yet ready to merge, still we have an issue with this fix.

lmignon:
@guewen, on our side @laetitia-gangloff is also testing these changes and that's why I don't have merged these changes despite the 3 thumbs up.

laetitia-gangloff:
  File "/usr/lib/python2.7/threading.py", line 810, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 763, in run
    self.__target(_self.__args, *_self.__kwargs)
  File "/home/lga/project/openerp/addons-connector/connector/jobrunner/runner.py", line 129, in urlopen
    set_job_pending()
  File "/home/lga/project/openerp/addons-connector/connector/jobrunner/runner.py", line 111, in set_job_pending
    "WHERE uuid=%s and state=%s", (PENDING, job_uuid, ENQUEUED)
ProgrammingError: column "worker_uuid" of relation "queue_job" does not exist
LINE 1: ...'pending', date_enqueued=NULL, date_started=NULL, worker_uui...

guewen:
:-1: until it is fixed

laetitia-gangloff:
I think ", worker_uuid=NULL" can be removed...
I did the following test :

start odoo
set a lot of job in the queue (7000)
start runner
wait 30 seconds
stop odoo
remove started job from my database
stop runner
start odoo
start runner

And all seems ok.

mdietrichc2c:
I have another issue: even when removeing the `autocommit` and adding `cr.commit()` at the end, the commit does not happen on an enqueued job (I created a timeout for testing).

connector/jobrunner/runner.py
```diff
@@ -98,19 +98,35 @@
...
+        url = ('http://localhost:%s/connector/runjob?db=%s&job_uuid=%s' %
+               (port, db_name, job_uuid))
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
             requests.get(url, timeout=1)
         except requests.Timeout:
-            pass
+            _logger.error("timeout in GET %s", url)
+            set_job_pending()
```
sbidoul: Timeout is a normal behaviour for long running jobs. So I would not log this as an error.

mdietrichc2c:
Turns out the error I had was due to my test (instant timeout => job was enqueued, set as pending, then running, enqueued, timeout...)

@sbidoul @laetitia-gangloff I corrected both points : removed the absent column from the SQL request , and the message is now logged as a `debug` message.

mdietrichc2c:
Actually, after discussing it with @guewen , every job longer than a second will log a timeout message ; it was decided to remove them for timeouts, and I rebased the whole thing in one commit.

guewen:
Thanks a lot @mdietrichc2c :-) 
I lift my thumb down :+1:

connector/jobrunner/runner.py
```diff
@@ -98,19 +98,34 @@
...
     def urlopen():
+        url = ('http://localhost:%s/connector/runjob?db=%s&job_uuid=%s' %
+               (port, db_name, job_uuid))
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
             requests.get(url, timeout=1)
         except requests.Timeout:
-            pass
+            set_job_pending()
```
lmignon, line 125: What will happen if a job is longer than 1 min? The state will be set to 'pending' and the running job will fails with a `concurrent update error`?

mdietrichc2c, line 125: Wouldn't the started job have `started` as a state, and hence not be reset as `pending` ?

guewen, line 125: Why? The state of the job is set to `started` when `/runjob` is called. From my understanding, this use case only cover a timeout that happen before we could hit the controller. If we hit the controller, the job is set to `started` so the job won't be reset to `pending`.

lmignon, line 125: @mdietrichc2c you're right

lmignon:
Thank you @mdietrichc2c . I can now comment my :+1:  with 'Code review and functional test'

laetitia-gangloff:
I redo the same test than before, and it works as expected :+1:

PR #151, opened 2015-11-19
Set job back to 'pending' in case of exception

guewen:
Forward-port of #150


lmignon:
as for #150 :+1: 
once/if travis is green

guewen:
Forward port accepted in #150. I merge.

PR #155, opened 2015-12-03
Fix the fix c3fefd2 :-(

guewen:
The first item contains the name of the DB such as 'odoo_db' and the
second item contains 'user=gbaconnier dbname=odoo_db' which is what
expects psycopg2.connect.

I feel stupid about my first correction especially since I was sure to have
correctly tested it (automated test is hard here). At least, it does not
prevent jobs to run.


pedrobaeza:
:+1:

lmignon:
:+1:  If you feel stupid, I share your feeling since I've approved the previous fix :smirk:

PR #158, opened 2015-12-07
[IMP] Allow the number of records being deleted to be limited to prev…

### LINKED ISSUE #157: [7.0] autovacuum times out or runs out of memory when deleting a huge amount of jobs
We have seen when a large number of jobs have built up the autovacuum job stops running correctly and the backlog of jobs to delete continues to increase. By adding an optional limit parameter and increasing the frequency of the autovacuum job we were able to get this backlog cleared eventually.

craiggowing:
…ent timeouts or out of memory issues

Fixes #157

### DIFF
--- a/connector/queue/model.py
+++ b/connector/queue/model.py
@@ -38,6 +38,7 @@ from ..connector import get_openerp_module, is_module_installed
 
 _logger = logging.getLogger(__name__)
 
+DEFAULT_LIMIT = 1000
 
 class QueueJob(orm.Model):
     """ Job status and result """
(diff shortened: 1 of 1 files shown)


lmignon:
@craiggowing Thank you for the proposal. IMO, It would be better to implement a kind of batch processing for the deletion. What about:

``` python

DEFAULT_LIMIT = 1000
...
  def autovacuum(self, cr, uid, limit=None, context=None):
         """ Delete all jobs (active or not) done since more than
         ``_removal_interval`` days.
         ``_removal_interval`` days limited by ``limit``.

         Called from a cron.
        :param limit: optional maximum number of records to delete at once
        :type limit: int
         """
         if context is None:
             context = {}
         deadline_fmt = deadline.strftime(DEFAULT_SERVER_DATETIME_FORMAT)
         while True:
             job_ids = self.search(cr, uid,
                                   [('date_done', '<=', deadline_fmt)],
                                  limit=limit or DEFAULT_LIMIT,
                                   context=context)
             if not job_ids:
                break;
             self.unlink(cr, uid, job_ids, context=context)
             cr.commit()
         return True
```

In this way we should limit the memory consumption and ensure that the timeout doesn't rollback more than one loop.

craiggowing:
@lmignon - that sounds like a good suggestion. It would then get round the need to change the frequency on the ir.cron record also.

guewen:
LGTM, thanks @craiggowing 
Could you squash your 2 commits?
:+1:

lmignon:
:+1:

craiggowing:
The commits are now squashed.

guewen:
Thanks!

Issue #58, opened 2015-04-22
worker is not supported dbfilter with %d or %h parameter.


ghost:
This connector worker is not supported with the dbfilter having either %d or %h.
i.e. dbfilter = '^%d.*' then it is not working with the following line of code.
dbfilter = config['dbfilter']
if dbfilter and db_names:
    db_names = [d for d in db_names if re.match(dbfilter, d)]

here dbfilter variable having '^%d.*' thats why re.match will not found any matches with this and it is returning [ ](blank list).

same this openerp-connector-worker script is not working with dbfilter.

guewen:
The `--db-filter` option is an option filtering what databases are shown on the web interface.
I'm not sure that it should restrict the job workers. The `--database` option seems more appropriate for this to me.

sbidoul:
I think the runner should honor `--db-filter`. For example, you may want a configuration with two odoo instances on the same machine, each serving a different set of databases controlled by `--db-filter`. The runner should listen to the exact same set of databases than the Odoo instance it's running in.

I did not know dbfilter could include %d. I thought I had copied the same db filtering algorithm as the Odoo server. To be verified.

guewen:
@sbidoul here is my interpretation:
`--db-filter` only restricts HTTP to serve databases which are not in the filter. It does not prevent for instance the crons to run on those databases. The option that really makes Odoo (including crons) run on a selection of databases is `--database`. So in your case, with 2 odoo on the same machine but different set of databases, you should probably configure a different `--database` option otherwise your cron will be mixed up between all the databases. I think the job runner should have the same behavior than the cron.

sbidoul:
@guewen It seems you are right indeed. I remember now where I got this part of code from: I copied it from the connector worker :wink: 

Still it's a bit strange to me. If you use `%h` and `%d` in the dbfilter, this means that you probably don't use `--database` and the Odoo instance sees databases served by other hosts. So the crons will run on all of them and compete to catch the jobs. That's probably a minor issues for crons which will lock the jobs anyway and prevent the same job to be run by two instance at the same time.

But for the runner, if it ignores dbfilter, it will listen (and create connections) to all visible databases.

So I would rather keep using dbfilter and use http.db_filter() in https://github.com/OCA/connector/blob/8.0/connector/jobrunner/runner.py#L221

flotho:
Hi folks,

I think I have this kind of issue. I run successfully the connector-worker yet no worker is added in the db that has the connector installed.
[...]

PR #159, opened 2015-12-08
Removing '%h' and '%b' from 'dbfilter' to avoid the wrong detection database (issue #58).

xt0ph:


connector/queue/worker.py
```diff
@@ -263,6 +263,8 @@ def available_db_names():
             db_names = db.exp_list(True)
         dbfilter = config['dbfilter']
         if dbfilter and db_names:
+            # Remove %d and %h from dbfilter (issue #58)
+            dbfilter = re.sub('(%d|%h).*$', '', dbfilter)
             db_names = [d for d in db_names if re.match(dbfilter, d)]
```
sbidoul, line 268: Does it make sense to match on dbfilter after removing %d and %h?

As @guewen noted in #58 it probably does not make sence to use dbfilter at all in the job runner.

So we could completely remove this filtering, or to be less brutal towards user who would depend on this, skip the filtering if dbfilter contains %d or %h.

xt0ph:
Except if the database name contains a fixed part referencing the instance. For example dbfilter contains the 'o80_%d', so I need the connector reference the database beginning with 'o80_'  but not all databases on the server.

sbidoul:
But if dbfilter is 'odoo_%d_80', then removing the %d makes no sense. So your situation works a bit by chance, no?

xt0ph:
Not by chance, but at the origin of the '%d' is part of the research, it will not run. We have a prefix on our bases so it works. This is not an ideal solution but it works much better than the original. If someone has a better idea for the filter works with the '%d' (in all cases) I'm interested.

sbidoul:
I now think the connector should completely ignore dbfilter. The very fact that a client hostname can be part of the filter proves it's not a concept to be used by the connecter which is completely independent from clients.

So as a first step I would write 

``` python
if dbfilter and '%d' not in dbfilter and '%h' not in dbfilter:
```

So it does not break with dbfilters containing %d or %h and it continues to work with other dbfilters until we deprecate it's use completely in the connector.

guewen:
Thanks @2cadz for the proposal.
I agree with @sbidoul. Please see #165 for the proposed correction path.

PR #165, opened 2015-12-27
[FIX] ignore dbfilter, fixes #58

### LINKED ISSUE #58: worker is not supported dbfilter with %d or %h parameter.

sbidoul:
dbfilter can contains %d or %h, referring to the client hostname.
Since connector does run independently of client, it can not rely
on the dbfilter concept.

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -112,7 +112,6 @@ Caveat
 from contextlib import closing
 import logging
 import os
-import re
 import select
 import threading
 import time
@@ -265,9 +264,6 @@ class ConnectorRunner(object):
             db_names = openerp.tools.config['db_name'].split(',')
         else:
             db_names = openerp.service.db.exp_list(True)
-        dbfilter = openerp.tools.config['dbfilter']
-        if dbfilter and '%d' not in dbfilter and '%h' not in dbfilter:
-            db_names = [d for d in db_names if re.match(dbfilter, d)]
         return db_names
 
     def close_databases(self, remove_jobs=True):
--- a/connector/queue/worker.py
+++ b/connector/queue/worker.py
@@ -19,7 +19,6 @@
 #
 ##############################################################################
 
-import re
 import logging
 import os
 import threading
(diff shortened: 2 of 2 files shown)


sbidoul:
Perhaps this one should wait 4.0 as it can change the behaviour of some installations relying on dbfilter. #166 could be used for 3.x.

pedrobaeza:
I don't think this can cause any side effect. Do you?

guewen:
It seems to me that if one uses `%d` or `%h`, it works by chance.

Anyway what we can do is:
1. Merge #166 in 8.0
2. Tag HEAD of 8.0 as 8.0.3.3.0
3. Merge #165, next planned release being 8.0.4.0.0

What do you think?

Thanks for the correction!

lmignon:
:+1: with the proposal of @guewen

sbidoul:
@guewen @lmignon this one to be included in the 9.0 version?

lmignon:
@sbidoul 9.0 has been rebased on the latest 8.0 2 hours ago and also include #165

guewen:
@lmignon the latest 8.0 contains #166 which is the backward compatible fix but not #165. I agree with @sbidoul to integrate this one in 9.0.

lmignon:
@guewen @sbidoul done https://github.com/acsone/connector/commit/bdf9bbfd4aefd8dbf3d1206e2855b84710f7d246

guewen:
thanks @lmignon !

sbidoul:
@lmignon that's an empty commit :wink: I had missed it but it was already in your PR at https://github.com/acsone/connector/commit/3c83de414638d955f4e302cb1f00a749b2bd5029.

So we can close this one as it is taken into account in #169

PR #166, opened 2015-12-27
[FIX] ignore dbfilter containing %d or %h, fixes #58

### LINKED ISSUE #58: worker is not supported dbfilter with %d or %h parameter.

sbidoul:
This is a temporary fix. In version 4.0, dbfilter should be completely ignored by connector.

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -266,7 +266,7 @@ class ConnectorRunner(object):
         else:
             db_names = openerp.service.db.exp_list()
         dbfilter = openerp.tools.config['dbfilter']
-        if dbfilter:
+        if dbfilter and '%d' not in dbfilter and '%h' not in dbfilter:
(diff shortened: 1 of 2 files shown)


pedrobaeza:
:+1: as a temporary patch

lmignon:
:+1:  OK for me for the temporary patch and and the proposal of @guewen to merge #165, in the next planned release being 8.0.4.0.0 (https://github.com/OCA/connector/pull/165#issuecomment-168605106)

guewen:
:+1: 
Thanks

PR #167, opened 2015-12-28
Manage non-ascii PG errors

npiganeau:
When PG is localized, error messages are not ascii and jobs are not postponed. Instead they are failed with a 'unicode decode error'.

Note: even if PG `lc_messages` is set to en_us.utf8 on a localized system, the error message prefixes (such as `DETAILS:`) are still localized.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -94,7 +94,8 @@ class RunJobController(http.Controller):
                 if err.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
                     raise
 
-                retry_postpone(job, unicode(err), seconds=PG_RETRY)
+                retry_postpone(job, unicode(err.pgerror, errors='replace'), 
+                               seconds=PG_RETRY)
                 _logger.debug('%s OperationalError, postponed', job)
 
         except NothingToDoJob as err:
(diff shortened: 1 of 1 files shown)


guewen:
:+1:

connector/controllers/main.py
```diff
@@ -94,7 +94,8 @@ def retry_postpone(job, message, seconds=None):
                 if err.pgcode not in PG_CONCURRENCY_ERRORS_TO_RETRY:
                     raise
 
-                retry_postpone(job, unicode(err), seconds=PG_RETRY)
+                retry_postpone(job, unicode(err.pgerror, errors='replace'),
```
lmignon: May be we can replace the call to `unicode` with 

``` python
from openerp import tools
...
    retry_postpone(job, tools.ustr(err), seconds=PG_RETRY)
...
```

https://github.com/odoo/odoo/blob/8.0/openerp/loglevels.py#L58

guewen:
I agree with @lmignon suggestion, could you change that? (possibly by passing `errors='ignore'` to `ustr` too).

guewen:
thanks @npiganeau
@lmignon ok for you?

lmignon:
:+1:  @guewen Ok for me!

PR #171, opened 2016-02-22
[7.0] Prevent to unpickle globals which are not jobs

guewen:
This is a safeguard to prevent someone to write arbitrary code in jobs.
Builtin types and datetime/timedelta are allowed in job arguments, and a
new function 'whitelist_unpickle_global' allows to register new objects
if needed.

I will do a proper release with a new addon's version and updated changelog in
a PR after this one is merged.

/cc @lmignon @sbidoul @gurneyalex @colinnewell

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -200,10 +200,11 @@ class Database(object):
             cr.execute("LISTEN connector")
 
     def select_jobs(self, where, args):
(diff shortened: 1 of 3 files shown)


lmignon:
:+1:  same as #170

colinnewell:
:+1:

lmignon:
Same changes as in #170 and 3 :+1:  on #170 -> merge

PR #172, opened 2016-03-02
Add job_uuid in context

YannickB:
Hello,

This PR add the job_uuid in the context of the function being executed. This allow to easily retrieve information which may be needed by the subfunction and relative to the job, like the date when the execution started, the job id in itself etc...

Context : We absolutely need it for the Clouder project, because the log of system calls are stored in the job itself. Currently we have no way to retrieve the job which called the function.

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -507,7 +507,8 @@ class Job(object):
         with session.change_user(self.user_id):
             self.retry += 1
             try:
-                self.result = self.func(session, *self.args, **self.kwargs)
+                with session.change_context({'job_uuid': self._uuid}):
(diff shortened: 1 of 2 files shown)


guewen:
Sound good. Could add a simple test such as this one https://github.com/OCA/connector/blob/8.0/connector/tests/test_job.py#L116-L119? The idea being a new test job that returns session.env.context.get('job_uuid').

lmignon:
Thank you @YannickB . :+1:

sebalix:
Useful :+1:

YannickB:
Thank you guys, the tests have been added by nicolas (and you should have received his CLA)

damdam-s:
:+1: thanks

PR #169, opened 2016-02-15
9.0 from 8.0

lmignon:
Migration to version 9.0
- [x] connector (@lmignon)
  - [x] forward port of latest changes from 8.0 to 9.0
  - [x] make the module installable (@lmignon)
  - [x] remove the old workers code (@lmignon)
  - [x] integrate PR #165 (@lmignon)

PR #178, opened 2016-03-03
Some more 9.0 migration

sbidoul:


guewen:
Thanks! Could you complete the changelog for 9.0.1.0.1?

sbidoul:
@guewen BTW, should we not version it 9.0.4.0.1 instead of 9.0.1.0.1?

People are used to talk version 3 of the connector, so it may be strange to fall back to 1.0.

pedrobaeza:
But version is 9.0.1.0.0, not only 1.0.0

lmignon:
I'm really surprised to see that the module is `'installable': False`. I can't figure out how this value has changed from https://github.com/acsone/connector/commit/29079562e2cf36cb33c12c4d068ca893d977977f. I've checked each following commit and this value has never been reset to `False`. This looks like a bug in git...
:+1:

sbidoul:
changelog updated

pedrobaeza:
:+1:

guewen:
> @guewen BTW, should we not version it 9.0.4.0.1 instead of 9.0.1.0.1?
> 
> People are used to talk version 3 of the connector, so it may be strange to fall back to 1.0.

Yeah.. well, previously we were only writing the last digits of the version, so if we consider only them it's weird. But I think we should speak about `8.0.3.3.0` and `9.0.1.0.1` and now that's fine. In fact, I didn't thought about that before reading https://github.com/OCA/connector/pull/169#discussion_r52924726 ;-)

Thanks for the changelog!

guewen:
@sbidoul @lmignon do you agree on the version numbering?

lmignon:
@guewen Yes I agree

sbidoul:
@guewen I don't disagree

guewen:
:+1:

PR #184, opened 2016-03-31
[REF] write function - more modularity

gfcapalbo:
Small refactoring of the write method in queue/model.py

I found my self in the necessity to write a module that silenced some error notifications that were bothering the client. I could not do a clean inheritance the way write() is structured now. I am proposing a refactoring to allow the write function to accept more future modularity.

Ideally to avoid the situation I found myself in we should have:

``` python
def write(self):
    self.pre_write()
    res=SUPER
    self.post_write()
    return res   



```

### DIFF
--- a/connector/queue/model.py
+++ b/connector/queue/model.py
@@ -148,8 +148,11 @@ class QueueJob(models.Model):
         return True
 
     @api.multi
-    def write(self, vals):
-        res = super(QueueJob, self).write(vals)
+    def pre_write(self):
+        pass
+
+    @api.multi
+    def post_write(self, vals):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
Please squash to one commit. Why not including finally the pre_write?

gfcapalbo:
Commits squashed.

guewen:
Could you show an example of what it allows to do that you couldn't do before?

lasley:
Thanks for the submission @gfcapalbo - I agree that a note on what this allows before/after would be excellent.

guewen:
Please re-open in case you still need it and have more details on the intent.

PR #185, opened 2016-04-04
[8.0] add date to whitelist_unpickle_global list

laetitia-gangloff:
I think date (as datetime and timedelta) can be allowed in whitelist unpickle.

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -24,7 +24,7 @@ import functools
 import logging
 import uuid
 import sys
-from datetime import datetime, timedelta, MINYEAR
+from datetime import date, datetime, timedelta, MINYEAR
 from cPickle import dumps, UnpicklingError, Unpickler
 from cStringIO import StringIO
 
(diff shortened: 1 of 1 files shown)


sebalix:
:+1: 

I encounter the `decimal.Decimal` type too (`import decimal`) while dealing with `pymssql`, is there an interest to add it?

laetitia-gangloff:
What is the best way to get a "complete" and "easy to extend" whitelist unpickle ?

guewen:
> What is the best way to get a "complete" and "easy to extend" whitelist unpickle ?

What do you mean?

guewen:
:+1:

guewen:
> I encounter the decimal.Decimal type too (import decimal) while dealing with pymssql, is there an interest to add it?

Not sure it is common enough to add it (unfortunately in Odoo, yes). You can add it in your addon with `whitelist_unpickle_global(decimal.Decimal)`

sebalix:
@guewen yes that's what I did, it is really not a problem as long as we can update this whitelist easily. Thanks!

sbidoul:
:+1:

lmignon:
:+1:

PR #187, opened 2016-04-07
8.0 manage invalid data error

gfcapalbo:
Problem:
I would have many jobs stuck in start.  THese jobs Had all raised InvalidDataError.
Strangely the Exception was not catched in main.py, but the generic 

Therefore the store function that had to write my failed job would never trigger

When explicitly importing InvalidDataError, the exception was catched (see commit  
https://github.com/gfcapalbo/connector/commit/5b77f3c6d77a9e10ee3f0f2e2fca72747e0f0244)

I replaced except (FailedJobError, InvaliDataError, Exception):  
with 
except:

I am still not sure why this works for me, but it seems OK to set ALL of the forseeable exceptions to fail jobs.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -15,8 +15,7 @@ from ..exception import (NoSuchJobError,
                          NotReadableJobError,
                          RetryableJobError,
                          FailedJobError,
-                         NothingToDoJob,
-                         InvalidDataError)
+                         NothingToDoJob)
 
 _logger = logging.getLogger(__name__)
 
@@ -112,7 +111,8 @@ class RunJobController(http.Controller):
             # delay the job later, requeue
             retry_postpone(job, unicode(err), seconds=err.seconds)
             _logger.debug('%s postponed', job)
-        except (FailedJobError, InvalidDataError, Exception):
+
+        except:
             buff = StringIO()
             traceback.print_exc(file=buff)
             _logger.error(buff.getvalue())
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -129,7 +129,6 @@ class RunJobController(http.Controller):
             buff = StringIO()
             traceback.print_exc(file=buff)
             _logger.error(buff.getvalue())
-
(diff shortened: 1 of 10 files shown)


gfcapalbo:
@StefanRijnhart   @hbrunn     what do you think?

hbrunn:
I think first of all, you need to propose to the correct target branch, which is 8.0. Then I think @sbidoul and @guewen are a lot better persons to ask for an opinion because to my knowledge, this repo is their baby

connector/controllers/main.py
```diff
@@ -112,11 +112,10 @@ def retry_postpone(job, message, seconds=None):
             retry_postpone(job, unicode(err), seconds=err.seconds)
             _logger.debug('%s postponed', job)
 
-        except (FailedJobError, Exception):
+        except:
```
sbidoul, line 115: AFAICS, `InvalidDataError` is a `ConnectorException`, which is `RuntimeError` which is an `Exception`, so changing this exception handler should not have any influence on your use case. Your root cause must be something else.

hbrunn, line 115: @sbidoul that's what I thought too. But the same applies to `FailedJobError`, and this one is mentioned explicitly here. `git blame` says you wrote this line, do you remember why you added `FailedJobError`? And do you agree in general to catch unconditionally here? I'd think we want to persist the failed state no matter what happened, right?

sbidoul, line 115: I was also wondering why the `FailedJobError` is there. This code comes from https://github.com/OCA/connector/blob/1dbde5d23dbd97c5fee4accd408cb619bfd36ce8/connector/queue/worker.py#L150 (@guewen's). 

Still, the current except statement is equivalent to `except Exception` which catches everything for all practical purposes (https://docs.python.org/2/library/exceptions.html#exceptions.Exception).

@gfcapalbo can you say more about your problem?

gfcapalbo:
Closed/
Please continue discussion here : Correct PR https://github.com/OCA/connector/pull/188

PR #191, opened 2016-04-29
Move connector settings menu to technical

lasley:
This PR moves the Connector settings menu to the Technical Settings submenu. 

Odoo 9 made the General Settings menu a link, which current implementation breaks by adding itself as a submenu.

### DIFF
--- a/connector/__openerp__.py
+++ b/connector/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 
 {'name': 'Connector',
- 'version': '9.0.1.0.2',
+ 'version': '9.0.1.0.3',
  'author': 'Camptocamp,Openerp Connector Core Editors,'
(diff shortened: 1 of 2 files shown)


lasley:
Hmmmm this docs error is something I haven't seen before - https://travis-ci.org/OCA/connector/jobs/126748870

Can anyone provide some guidance on that? Is this part of the new setuptools build process?

guewen:
Thanks! :+1: 

The doc build issue comes from https://github.com/OCA/odoo-sphinx-autodoc/pull/4 which has been fixed but not yet pushed on pypi.

lasley:
Nice, I'm going to have to check out this AutoDoc module - it looks handy! Thanks for the clarification & review

lmignon:
:+1:

PR #153, opened 2015-12-01
Fix backend_to_m2o to extract id of the binding

guewen:
The default binder returns a recordset of one record since 41561c8 so
backend_to_m2o should extract the Id from it.

PR #194, opened 2016-05-12
Fix backend_to_m2o to extract id of the binding

guewen:
The default binder returns a recordset of one record since 41561c8 so
backend_to_m2o should extract the Id from it.

Port from https://github.com/OCA/connector/pull/153


lmignon:
Same as for #153 :+1: LGTM (Code review no test)

lasley:
👍  waiting on rebuild for 🍏

PR #201, opened 2016-08-10
Porting job_uuid in context to V9

nicolas-petit:
See the following pull request for V8 version:
https://github.com/OCA/connector/pull/172/files

### DIFF
--- a/connector/queue/job.py
+++ b/connector/queue/job.py
@@ -486,7 +486,8 @@ class Job(object):
         with session.change_user(self.user_id):
             self.retry += 1
             try:
-                self.result = self.func(session, *self.args, **self.kwargs)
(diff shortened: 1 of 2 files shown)


lasley:
Hi @nicolas-petit - thanks for the submission. Would it be possible to instead cherry pick the relevant commits from the 8.0 branch? While I admittedly prefer the one commit that we have here vs the multiple there, using the cherry-picks makes it easier to identify the change sets are actually the same from a high level.

nicolas-petit:
Thank your for your feedback @lasley 
I revised this pull request using cherry-picking.

lmignon:
:+1: (Code review)

sebalix:
:+1:

lasley:
Thanks for the update @nicolas-petit 👍

PR #205, opened 2016-08-22
[9.0] delay between job start, configurable per channel

### LINKED ISSUE #203: Blueprint: per channel delay between jobs
# Goal

The current channel configuration mechanism allows to limit the number of jobs that can run in parallel. However, as soon as there is available capacity, the next job runs immediately (provided there is capacity in the downstream channel, of course). 

Some third party systems have rate limitation systems that reject requests that arrive with a too high frequency. 

To cope with such systems, I propose to extend the channel configuration mechanism to implement a delay betwen the dequeing of successive jobs.
# Configuration

The channel configuration string will accept a new key 'delay' whose value is the delay between jobs in seconds.

https://github.com/OCA/connector/blob/0df11a02610c04cd600ce087ff9ba274113c403d/connector/jobrunner/channels.py#L539
# Implementation

It should be as simple as using this delay option to compute the time before which no job can be dequed in the channel, then breaking the get_jobs_to_run loop if that time is not reached.

https://github.com/OCA/connector/blob/0df11a02610c04cd600ce087ff9ba274113c403d/connector/jobrunner/channels.py#L460

sbidoul:
fixes #203

### DIFF
--- a/connector/doc/guides/jobrunner.rst
+++ b/connector/doc/guides/jobrunner.rst
@@ -25,7 +25,8 @@ Intermediate subchannels which are not configured explicitly are autocreated
 with an unlimited capacity (except the root channel which if not configured gets
 a default capacity of 1).
 
-A delay between jobs can be set at the channel level with the ``delay`` key.
+A delay in seconds between jobs can be set at the channel level with 
+the ``throttle`` key.
 
 Example ``ODOO_CONNECTOR_CHANNELS``:
 
@@ -33,4 +34,4 @@ Example ``ODOO_CONNECTOR_CHANNELS``:
 * ``root:4,root.sub:2``: allow up to 4 concurrent jobs in the root channel and
   up to 2 concurrent jobs in the channel named ``root.sub``.
 * ``sub:2``: the same.
-* ``root:4:delay=2``: wait at least 2 seconds before starting the next job
+* ``root:4:throttle=2``: wait at least 2 seconds before starting the next job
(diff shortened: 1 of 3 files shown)


guewen:
Amazing work :+1:

lmignon:
Once again, I am impressed by such a so clean implementation!! :+1: (Core review, no tests)

connector/doc/guides/jobrunner.rst
```diff
@@ -25,9 +25,12 @@ Intermediate subchannels which are not configured explicitly are autocreated
 with an unlimited capacity (except the root channel which if not configured gets
 a default capacity of 1).
 
+A delay between jobs can be set at the channel level with the ``delay`` key.
```
pedrobaeza: A delay **in seconds**...

guewen: Just wondering if `wait` could not be more appropriate, since `delay` is also the name of the method to delay a job?

sbidoul: Yep, that traversed my mind too. I was also considering `pause` or `throttle`. I personally like 
`throttle`. Any preference?

guewen: I like `throttle`, it seems the most accurate term to me.

lmignon: :+1:  for `throttle`

sbidoul:
`throttle` key implemented and doc updated.

guewen:
Thanks

lasley:
Gotta love it when the comments are deeper than the code! 👍

PR #207, opened 2016-09-06
Fix wrong recordset returned when no binding found

guewen:
When we call the binder 'to_openerp' method with 'unwrap', we expect the
binder to return the 'unwrapped' record of the binding, e.g.
product.product instead of prestashop.product.product.

When no record is found, the binder _must_ return an empty recordset of
the correct (unwrapped) model too.


sebalix:
:+1:

lasley:
Definitely 👍

lmignon:
:+1:

atchuthan:
:+1:

PR #208, opened 2016-09-14
[9.0][FIX] connector: Zero value assertion

lasley:
- Switch zero assertion to type strict eval (`is`)

Before:

```
>>> external_id = False
>>> binding_id = True
>>> assert (external_id or external_id == 0) and binding_id
# No assertion raised
```

After

``` python
>>> external_id = 'test'
>>> assert (external_id or external_id is 0) and binding_id
>>> external_id = 0
>>> assert (external_id or external_id is 0) and binding_id
>>> external_id = False
>>> assert (external_id or external_id is 0) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
>>> external_id = None
>>> assert (external_id or external_id is 0) and binding_id
  File "<stdin>", line 1, in <module>
AssertionError
```

Taken from my fix in OCA/connector-magento#220 - not sure why the code is duplicated ¯_(ツ)_/¯

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -416,7 +416,7 @@ class Binder(ConnectorUnit):
         :type binding_id: int
         """
         # Prevent False, None, or "", but not 0
-        assert (external_id or external_id == 0) and binding_id, (
+        assert (external_id or external_id is 0) and binding_id, (
             "external_id or binding_id missing, "
             "got: %s, %s" % (external_id, binding_id)
         )


guewen:
The code is duplicated because there was no implementation of the Binder in the connector's core, the implementation was let in charge of the sub-addons. A default Binder has been added in https://github.com/OCA/connector/pull/91 so it should be removed from the Magento implementation.

:+1:

pedrobaeza:
:+1:

simahawk:
:+1:

lasley:
Thanks for the elaboration @guewen - now I feel less bad about this being duplicated in all my connectors too 😆

PR #210, opened 2016-09-28
[IMP] Allow configuring runner from config file

leorochael:
In addition to environment variables.

Environment variables are useful in the case where all odoo configuration is done through command line arguments, since we can't augment the list of command line switches from a module.

But in case odoo is started with a configuration file, allow it to declare the channels configuration, thereby enabling the runner.

This still currentlyl requires passing `--load=web,web_kanban,connector` on the command line, since there is no way to pass `load` by configuration file, but I think it's already an improvement.


sbidoul:
@leorochael this is intriguing. I was sure to have tested that when initially writing jobrunner, but I did not succeed. But it is very possible I was mixing the enabling of the runner and the configuration of  channels.

If it works it would be great.

The only change I would suggest is having a dedicated `[connector]` section in the config file to open the door to future extensions.

leorochael:
@sbidoul, I changed it to read info from a section named `[options-connector]`, as that can be generated from [`anybox.recipe.odoo`](http://pythonhosted.org/anybox.recipe.odoo/) using a syntax like:

```
[odoo]
recipe = anybox.recipe.odoo:server
(...)
options-connector.channels = root:4
```

Being able to generate a working configuration with `anybox.recipe.odo` without resorting to environment variables is a big motivation for this PR.

Anything else needed before this is merged?

sbidoul:
@leorochael I'll do some tests as soon as possible.

I'm not sure which one must have priority: the config file or the env var.

The documentation would need an update too.

leorochael:
@sbidoul: Can you give me some pointers on how to update the docs?

leorochael:
And I would say the env var should have priority, now that you mention it, as it's more "malleable" than a config file.

I can update it quickly if you want.

sbidoul:
@leorochael the doc in the `.rst` files, for example https://github.com/OCA/connector/blob/9.0/connector/doc/guides/jobrunner.rst

leorochael:
@sbidoul, Ok, as soon as you're satisfied with the code, let me know and I'll update the docs.

connector/jobrunner/__init__.py
```diff
@@ -44,9 +44,16 @@
...
+# On the other hand, the odoo configuration file can be extended at will,
+# so we check it in addition to the environment variables.
 
 
-enable = os.environ.get('ODOO_CONNECTOR_CHANNELS')
+def _channels():
+    return (
+        os.environ.get('ODOO_CONNECTOR_CHANNELS') or
+        config.misc.get("options-connector", {}).get("channels")
+    )
```
sbidoul: I think the config file must have precedence over the environment variable as the config is more specific to a running instance.

sbidoul: As a section name, `[options-connector]` is not as nice as `[connector]`, but if it's the only way to have it work with buildout, it's fine with me.

sbidoul:
@leorochael The environment variable is also used in https://github.com/OCA/connector/blob/b2658d448c9ed3ffbd72aafc704002bf71ab941c/connector/queue/worker.py#L353 to disable the old wokers when the runner is enabled. You'll need to handle that too. Note in v9 the runner is always enabled and the workers are removed, but in 8.0 we must preserve backward compatibility so it's the worker which run by default.

leorochael:
@sbidoul, I've handled worker and updated the documentation. See if it's to your liking.

connector/doc/guides/jobrunner.rst
```diff
@@ -19,13 +19,20 @@ The ``ODOO_CONNECTOR_CHANNELS`` environment variable must be
 set before starting Odoo in order to enable the job runner
 and configure the capacity of the channels.
 
+Alternatively, set the cannel configuration in the Odoo configuration file:
```
guewen: small typo: `s/cannel/channel/`

connector/doc/guides/jobrunner.rst
```diff
@@ -19,13 +19,20 @@ The ``ODOO_CONNECTOR_CHANNELS`` environment variable must be
...
+    channels = root:4
+
 The general syntax is ``channel(.subchannel)*(:capacity(:key(=value)?)*)?,...``.
 
 Intermediate subchannels which are not configured explicitly are autocreated
 with an unlimited capacity (except the root channel which if not configured gets
 a default capacity of 1).
 
-Example ``ODOO_CONNECTOR_CHANNELS``:
+Example values for ``ODOO_CONNECTOR_CHANNELS`` or ``options-connector:cannels``:
```
guewen: small typo: `s/cannel/channel/`

guewen (commented):
This is great! Thanks

Just small typos in the documentation and we are good, if @sbidoul agrees.

connector/jobrunner/runner.py
```diff
@@ -131,6 +135,21 @@
...
+# to configure the runner (channels mostly).
+#
+# On the other hand, the odoo configuration file can be extended at will,
+# so we check it in addition to the environment variables.
+
+
+def _channels():
+    return (
+        os.environ.get('ODOO_CONNECTOR_CHANNELS') or
+        config.misc.get("options-connector", {}).get("channels")
```
sbidoul: I suggest to swap these two lines to give precedence to the config file.

My reasoning is that the config file is more specific and more visible than the environment variable so it is less surprising to have the config have precedence over the env var.

leorochael:
@guewen, embarassing typos fixed. Took the opportunity to add highlight to the config example.

@sbidoul, if this is acceptable, I'll start working on a forward-port to branches 9 and 10 as soon as this is merged.

guewen (approved):
Thanks!

connector/doc/guides/jobrunner.rst
```diff
@@ -19,13 +19,20 @@ The ``ODOO_CONNECTOR_CHANNELS`` environment variable must be
 set before starting Odoo in order to enable the job runner
 and configure the capacity of the channels.
 
+Alternatively, set the channel configuration in the Odoo configuration file:
+
+.. code-block:: cfg
+
+    [options-connector]
+    channels = root:4
```
pedrobaeza, line 27: It would be easier if we can define channels this way:

```
[options-connector]
channels: 
  root: 4,
  root.connector-magento: 2
...
```

(I think they are standard configuration files options)

Is that possible?

leorochael: Hi @pedrobaeza,

To make your proposed syntax work I had to add whitespace handling to the channel configuration parsing. Should be backward compatible.

So yes, your proposed syntax should work now, and it even allows for trailing spaces and commented lines (which are discarded by the config parser and manifest as empty entries).

sbidoul (approved):

sbidoul (approved):

pedrobaeza (approved):
Once you change the precedence

connector/jobrunner/runner.py
leorochael: @sbidoul, you're right that the config file is more visible and more specific than env variables.

On the other hand, env variables are easier to change between invocations than config file values, as you can do it directly on the command line like:

```
ODOO_CONNECTOR_CHANNELS=subchannel:5 bin/start_odoo
```

Also, some deployment systems and PaaS services make it easy to use the same setup (i.e. the same set of code and configuration) in different roles by changing environment variables.

Then again, if someone wants to use it like this, I guess they could just leave the config file empty...

I'd just like to point out that some systems use environment variables to override settings in the configuration file. One notable example is [git](https://git-scm.com/docs/git-config), where things like `user.name` and `user.email` from the configuration can conveniently be changed for the scope of a single commit by setting `GIT_AUTHOR_NAME` and `GIT_AUTHOR_EMAIL`.

In any case, @sbidoul, @pedrobaeza, I'm not dead-set against changing the priority. I'd just like to ask you to consider the above points one more time and give me a final opinion.

If you decide that environment should have priority after all, I'll add one final change to the code above to specify that, if the env var is set, even if it's empty, then it will cancel the channel configuration from the file, so that it's possible to disable the runner with an invocation like.

```
ODOO_CONNECTOR_CHANNELS="" bin/start_odoo
```

Otherwise I'll swap things around.

guewen: I agree with @leorochael, we might want to deploy an instance with a configuration file, which is kind of fixed, but just override the configuration for one invocation using the env variable

pedrobaeza:
Yeah, good point. For me we can keep it as is then.

sbidoul:
Tools that give precedence to env vars (such git or pip) have a kind of per user config file and don't accept the config file as a command line option.

Odoo has both and we can't differentiate between `~/.odoorc` and `-c odoo.cfg` so its not so clear-cut.

In the end I don't think it is that important.

connector/doc/guides/jobrunner.rst
leorochael: I've improved the whitespace handling further by allowing to replace the commas with line breaks, which allows for even prettier configuration of channels in the file.

leorochael:
Ok, so for now the balance of opinions is tipped in the direction of env vars overriding conf files, so I just incremented it a bit to allow disabling the configuration file channels with the environment.

I also incremented a bit the whitespace handling to treat line breaks the same as commas, and clarified it in the documentation.

If you'd like further changes, let me know.

sebalix:
Nice feature! I also agree to give the priority to env vars.

sebalix (approved):

lasley (approved):
Great feature indeed, also agreed on ENV var precedence 👍

leorochael:
Thanks everyone!

Now if you could just weight in on [this Odoo PR](https://github.com/odoo/odoo/pull/13685), we may be able to get to a state where we can launch the runner with no command line switches whatsoever...

By the way, should I resubmit (an adaptation of) this merged PR to the 9.0 and 10.0 branches as well? What is the procedure for forward porting these things?

pedrobaeza:
Yes, it would be great to be proposed also on the rest of the versions.

leorochael:
I've also proposed the odoo configuration file change for site-wide modules here OCA/OCB#553, while it's still pending on the Odoo side.

I also ported this change to the 9.0 branch as #213, where it's slightly different than the 8.0 implementation.

Your comments on both PRs would be much appreciated...

PR #213, opened 2016-10-06
[IMP] [9.0] Allow configuring runner from config file

leorochael:
A forward port of #210.

It is slightly different from the 8.0 version since,  AFAICS, the 9.0 connector always runs with a `root:1` channel configuration if the `connector` module is `--load`ed server wide on Odoo and the environment variable is missing or set to empty.

Which means the only way to "disable" the `jobrunner` is not to load the `connector` module at all. Which is fine IMO.

So I moved the default channel configuration logic inside the Channel runner itself.

This could be considered a slightly backward incompatible change:
- Previously, if the `ChannelRunner` was not invoked by `--load`ing the `connector` module server wide, and someone decided to manually invoke the `ChannelRunner` without passing a channel configuration: 
  - then, the default channel configuration would be `root:1`
  - and not whatever was specified by `os.env['ODOO_CONNECTOR_CHANNELS']`.

With this PR, if the `ChannelRunner` is invoked manually without passing a channel configuration, then instead of defaulting directly to `root:1`, the environment variable and the configuration file will be consulted instead, defaulting to `root:1` only if both are missing.


leorochael:
Any opinion on this? What is necessary to move this forward?

pedrobaeza (approved):

pedrobaeza:
@guewen @sbidoul can you review this?

sebalix (approved):

connector/jobrunner/runner.py
```diff
@@ -119,6 +134,22 @@
 _logger = logging.getLogger(__name__)
 
 
+# Unfortunately, it is not possible to extend the Odoo
```
lasley, line 137: Should this maybe be a DocBlock for `_channels`?

lasley (approved):
Minor comment, non-blocking. Thanks for the forward porting the improvement 👍

connector/jobrunner/channels.py
```diff
@@ -632,12 +632,21 @@ def __init__(self):
...
         self._channels_by_name = WeakValueDictionary(root=self._root_channel)
 
+    @staticmethod
+    def split_strip(s, sep, maxsplit=-1):
+        """Split string and strip each component.
+
+        >>> ChannelManager.split_strip("foo: bar baz\\n: fred:", ":")
+        ['foo', 'bar baz', 'fred', '']
+        """
+        return [x.strip() for x in s.split(sep, maxsplit)]
```
sbidoul, line 642: This is more a candidate for a simple function rather than a class method.

sbidoul (approved):
Minor remark, other wise looks very good to me (code review, no test). Thanks for this work and care to update the documentation.

connector/jobrunner/runner.py
leorochael: Perhaps... The first part of this comment was moved (and refrased) from `jobrunner/__init__.py`, so it didn't occur to me to change it into a docstring.

If I have to shave this yak some more due to @sbidoul's comments before merging this branch I'll take the opportunity to move this comment into a `_channels()` docstring as well. :smirk:

connector/jobrunner/channels.py
leorochael: It was already like this on the 8.0-branch version of this change.

I could easily fix this here, but I'm afraid it could impact future changes done to multiple branches.

Should I:
- fix it just here (and on the future branch 10 version of this PR)?
- leave it as is?
  - And perhaps in the future do another PR for fixing this in the three branches?

sbidoul: @leorochael I approved the PR already. So you can leave as is.

PR #214, opened 2016-10-11
[FIX] Prevent MissingError to be raised due to dangling recomputations

StefanRijnhart:
Fixes jobs getting stuck in 'started' state after an exception was raised during the execution of a job in which a record was created.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -80,6 +80,11 @@ class RunJobController(http.Controller):
                 job.set_pending(reset_retry=False)
                 self.job_storage_class(session).store(job)
 
+        def clear_env(env):
+            """ Clear any dangling recomputations from failed job """
+            env.clear_recompute_old()
+            env.all.todo.clear()
+
         with session_hdl.session() as session:
             job = self._load_job(session, job_uuid)
             if job is None:
@@ -105,6 +110,7 @@ class RunJobController(http.Controller):
                 msg = None
             job.cancel(msg)
             with session_hdl.session() as session:
+                clear_env(session.env)
                 self.job_storage_class(session).store(job)
 
         except RetryableJobError as err:
@@ -119,6 +125,7 @@ class RunJobController(http.Controller):
 
             job.set_failed(exc_info=buff.getvalue())
             with session_hdl.session() as session:
+                clear_env(session.env)
                 self.job_storage_class(session).store(job)
             raise
 


sbidoul:
@StefanRijnhart that's weird. I thought session_hdl.session() would create a new transaction with a new environment and all. 

https://github.com/OCA/connector/blob/95b80647c0c51968e4d0436bf8c64619eb10349d/connector/session.py#L75

So I don't understand how clearing the environment at that point helps, since it looks like it is a fresh new one already

lmignon (approved):
It makes sense (Code review only)

lmignon:
@sbidoul I think that the problem comes from the fact that `with openerp.api.Environment.manage():` can reuse an environment already created https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L741
But I may be wrong....

StefanRijnhart:
Apart from environment recycling, the todos are gathered from all the environments of the class https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L967

sbidoul:
@lmignon @StefanRijnhart so it is an Odoo bug that fails to cleanup the environment before recycling it?

StefanRijnhart:
No, it is by design so that the TODOs are consistent across environments.

sbidoul:
Even after a rollback?

StefanRijnhart:
Continuing after a rollback is not supported by the framework. As always, when doing fancy stuff with the cursor (cq. the environment) you have to know what you are doing and clean up after yourself.

sbidoul:
@StefanRijnhart okay. Not intuitive (to me).

One last question, would it reduce the surprise factor if you put the env cleaning in the ConnectorSessionHandler here: https://github.com/OCA/connector/blob/6b954bf8cfa1fd213f720606651e37b7c081b077/connector/session.py#L85?

StefanRijnhart:
Ah, now it gets interesting. Four lines above the one you linked[1], a core Odoo decorator clear_upon_failure is applied[2]. This in turn calls env.clear()[3], which amongst other things clears the new API todos[3], but not the old API todos[4]. And the problem in my case consisted of remaining old API todos. So maybe the bug is that this method should do that as well, and then this is an Odoo bug indeed.

[1] https://github.com/OCA/connector/blob/6b954bf8cfa1fd213f720606651e37b7c081b077/connector/session.py#L80
[2] https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L880
[3] https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L872
[4] https://github.com/odoo/odoo/blob/8.0/openerp/api.py#L970

StefanRijnhart:
BTW that decorator is not used anywhere in Odoo.

guewen:
I tracked down a similar issue here: https://github.com/odoo/odoo/issues/9236

Cleaning the old API todos as well would seem legit.

lmignon:
@StefanRijnhart Don't forget to increment the version number into the manifest. `'8.0.3.3.0'` -> `'8.0.3.3.1'`,

sbidoul:
Actually I tend to think the jobs manipulations in the controller should be done without the orm. It brings more trouble than benefits for such simple manipulations. cc/ @adrienpeiffer

PR #215, opened 2016-10-12
Fix some tests

simahawk:
1. currently the test `test_default_binder` is not tested at all on runbot
2. fix test_default_binder
3. make test_job work with nose

Unfortunately doctests loader in test_runner_channels.py do not work with nose, it throws this error:

```
ERROR: connector.tests.test_runner_channels.load_tests
----------------------------------------------------------------------
  File "/path/to/nose-1.3.7-py2.7.egg/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: load_tests() takes exactly 3 arguments (0 given)
```

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -449,10 +449,10 @@ class Binder(ConnectorUnit):
         else:
             binding = self.model.browse(binding_id)
 
-        openerp_record = getattr(binding, self._openerp_field)
+        record = getattr(binding, self._openerp_field)
         if browse:
-            return openerp_record
-        return openerp_record.id
+            return record
+        return record.id
 
     def unwrap_model(self):
         """ For a binding model, gives the normal model.
--- a/connector/tests/__init__.py
+++ b/connector/tests/__init__.py
@@ -30,3 +30,4 @@ from . import test_related_action
 from . import test_runner_channels
 from . import test_runner_runner
 from . import test_checkpoint
+from . import test_default_binder
--- a/connector/tests/test_default_binder.py
+++ b/connector/tests/test_default_binder.py
@@ -37,19 +37,28 @@ class TestDefaultBinder(TransactionCase):
         partner.write({'color': 1})
         # bind the main partner to external id = 0
         self.partner_binder.bind(0, partner.id)
+
         # find the openerp partner bound to external partner 0
-        openerp_id = self.partner_binder.to_openerp(0)
-        self.assertEqual(openerp_id, partner.id)
-        openerp_id = self.partner_binder.to_openerp(0)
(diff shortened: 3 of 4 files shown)


connector/connector.py
```diff
@@ -443,10 +443,11 @@ def unwrap_binding(self, binding_id, browse=False):
         else:
             binding = self.model.browse(binding_id)
 
-        openerp_record = getattr(binding, self._openerp_field)
+        # XXX: do we expect to have a m2o field here?
```
simahawk: @guewen this is not clear to me

guewen: `self._openerp_field` should be the m2o fields towards the "normal" record (the m2o used in the `_inherits`), does it answer to you question?

simahawk: fixed the test as we said ;)

connector/connector.py
```diff
@@ -443,10 +443,10 @@ def unwrap_binding(self, binding_id, browse=False):
         else:
             binding = self.model.browse(binding_id)
 
-        openerp_record = getattr(binding, self._openerp_field)
+        record = getattr(binding, self._openerp_field)
```
lasley, line 452: [This is a security issue](https://github.com/odoo/odoo/issues/11087). Does [`binding[self._openerp_field]` work](https://github.com/odoo/odoo/commit/709323f071cc9df27bf11a8cd216b329cc3045af#diff-01d4c5da8bcc205ddb9cad8c573a49e7L163)?

lasley (changes requested):
One note + please rebase & resolve conflict

connector/connector.py
lmignon: @lasley IMO it's not a security issue since `_openerp_field` is defined in the source code and not exposed to the ui...

lasley: @lmignon - Seems like a best practice to use item lookup instead of attribute for recordsets IMO, but you are correct that it's not editable in the UI. I'll remove the PR block

lasley (approved):

connector/connector.py
simahawk: @lasley @lmignon actually I planned to go trough all these `getattr` (there are several of them) and replace them, but as Laurent pointed out there's no user input so I think that we can live with this for a while :)

simahawk: for the record https://github.com/OCA/connector/issues/218 :)

lmignon: @lasley You are right, it's a best practice :smirk:

lmignon:
@simahawk Can you rebase and resolve the conflict?

simahawk:
@lmignon should be done :)

lmignon (approved):

PR #216, opened 2016-10-12
Improve checkpoint

simahawk:
- add checkpoint tests
- add `add_checkpoint_message` for just displaying a message without any related object
- a record checkpoint can have a message too

Note, this is based on https://github.com/OCA/connector/pull/215

### DIFF
--- a/connector/checkpoint/checkpoint.py
+++ b/connector/checkpoint/checkpoint.py
@@ -48,6 +48,8 @@ class ConnectorCheckpoint(models.Model):
     @api.depends('model_id', 'record_id')
     def _compute_record(self):
         for item in self:
+            if not (item.model_id and item.record_id):
+                continue
             item.record = item.model_id.model + ',' + str(item.record_id)
 
     @api.depends('model_id', 'record_id')
@@ -128,7 +130,7 @@ class ConnectorCheckpoint(models.Model):
     _sql_constraints = [
         ('required_fields',
          "CHECK (record_id IS NOT NULL OR message IS NOT NULL)",
-         _("Provide relation to a record or a message!")),
+         _("Provide relation to a record or a message.")),
     ]
 
     @api.multi
--- a/connector/checkpoint/checkpoint.py
+++ b/connector/checkpoint/checkpoint.py
@@ -49,6 +49,7 @@ class ConnectorCheckpoint(models.Model):
     def _compute_record(self):
         for item in self:
             if not (item.model_id and item.record_id):
+                item.record = None
                 continue
             item.record = item.model_id.model + ',' + str(item.record_id)
 
--- a/connector/checkpoint/checkpoint.py
+++ b/connector/checkpoint/checkpoint.py
@@ -159,14 +159,15 @@ class ConnectorCheckpoint(models.Model):
 
     @api.model
     def create_from_name(self, model_name, record_id,
-                         backend_model_name, backend_id):
+                         backend_model_name, backend_id, message=''):
         model_model = self.env['ir.model']
         model = model_model.search([('model', '=', model_name)], limit=1)
         assert model, "The model %s does not exist" % model_name
         backend = backend_model_name + ',' + str(backend_id)
         return self.create({'model_id': model.id,
                             'record_id': record_id,
-                            'backend_id': backend})
+                            'backend_id': backend,
+                            'message': message})
 
     @api.model
     def create_from_message(self, backend_model_name, backend_id, message):
@@ -183,10 +184,11 @@ class ConnectorCheckpoint(models.Model):
 
 
 def add_checkpoint(session, model_name, record_id,
-                   backend_model_name, backend_id):
+                   backend_model_name, backend_id, message=''):
     checkpoint_model = session.env['connector.checkpoint']
     return checkpoint_model.create_from_name(model_name, record_id,
-                                             backend_model_name, backend_id)
+                                             backend_model_name, backend_id,
+                                             message=message)
 
 
 def add_checkpoint_message(session, backend_model_name, backend_id, message):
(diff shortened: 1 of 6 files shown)


connector/checkpoint/checkpoint.py
```diff
@@ -47,14 +47,17 @@ def _reference_models(self):
 
     @api.depends('model_id', 'record_id')
     def _compute_record(self):
-        for check in self:
-            check.record = check.model_id.model + ',' + str(check.record_id)
+        for item in self:
```
guewen, line 50: I think you should add

``` python
if not (item.model_id and item.record_id):
    continue
```

connector/checkpoint/checkpoint.py
```diff
@@ -115,6 +118,18 @@ def _search_record(self, operator, value):
...
+        string='Message',
+        help="Review message",
+        readonly=True,
+        required=False,
+    )
+
+    _sql_constraints = [
+        ('required_fields',
+         "CHECK (record_id IS NOT NULL OR message IS NOT NULL)",
+         _("Provide relation to a record or a message!")),
```
guewen: Could you remove the `!` which is a bit too emphatic?

connector/connector.py
```diff
@@ -443,10 +443,11 @@ def unwrap_binding(self, binding_id, browse=False):
...
             binding = self.model.browse(binding_id)
 
-        openerp_record = getattr(binding, self._openerp_field)
+        # XXX: do we expect to have a m2o field here?
+        openerp_id = getattr(binding, self._openerp_field)
         if browse:
-            return openerp_record
-        return openerp_record.id
+            return self.model.browse(openerp_id)
+        return openerp_id
```
guewen: this changes comes from #215

connector/doc/guides/bootstrap_connector.rst
```diff
@@ -208,6 +211,23 @@ created. I propose to create a helper too in
         return checkpoint.add_checkpoint(session, model_name, record_id,
                                          'coffee.backend', backend_id)
 
+Message checkpoint
+------------------
+
+When you need to show a plain text message to a user
+(for instance: an import job did not succeeded),
```
guewen: I'm not sure than "an import job did not succeeded" is a good example: if a job really failed, its state will be failed and we can't miss it, we don't need to create a checkpoint. 

I was looking for a better example but couldn't find.

For instance: "A product has been imported but could not be attached to the proper category", but in that case you would link the checkpoint to the product AND show the message, don't you think?

guewen (changes requested):
Nice feature, thanks

connector/connector.py
simahawk: removed

connector/checkpoint/checkpoint.py
simahawk: done

connector/checkpoint/checkpoint.py
simahawk: done

connector/doc/guides/bootstrap_connector.rst
simahawk: Indeed, the issue we were thinking about that led us to this approach was: "I want to display a nice message to the end user instead of an error traceback, whenever I don't have an object".

Imagine you are importing a product and the product is not in odoo yet, and for some reason it fails (maybe because you put an hard constraint in the import chain). At this point you do not have any object to attach the checkpoint to and the only way to see this is to go and search for failed jobs.

This could be the good example, I guess :)

guewen: But go and search in the failed job should not be an issue, usually you should have _no_ failed jobs, as a user you have to monitor them. Besides, you can subscribe to the job failures to receive notifications. The use case I saw with you in the prestashop connector was more like the one I described: you are importing a product, but you couldn't import the category associated to the product. Never mind, we mark the job as success and add a checkpoint saying that its category could not be imported. I didn't realized before now that in this case we have a model to attach to and that's the product. Anyway, I'm sure there are use cases where adding a message without a record is needed.

simahawk: Another example: you are importing in batch mode partners via CSV. If you make the whole job fail for just one missing partner is way too much, right?
In this case you can simply add a checkpoint and tell the user to check the line N in the csv file.

guewen: This one is a good example. This is more for warnings in fact.

simahawk: @guewen there you go https://github.com/OCA/connector/pull/216/commits/588b598369c529db08b7b9f52ff4ae4c98175f16

guewen (approved):
Thanks for the updates!

lasley (approved):
Woot I like 👍  code, no test

simahawk:
@guewen now a record checkpoint can have a message, so we can handle better situations like this https://github.com/guewen/connector-prestashop/blob/9.0-migration-connector-prestashop/connector_prestashop/models/res_partner/importer.py#L218

guewen:
Nice

connector/checkpoint/checkpoint.py
```diff
@@ -47,14 +47,19 @@ def _reference_models(self):
 
     @api.depends('model_id', 'record_id')
     def _compute_record(self):
-        for check in self:
-            check.record = check.model_id.model + ',' + str(check.record_id)
+        for item in self:
+            if not (item.model_id and item.record_id):
+                continue
```
lmignon, line 53:
``` python
    if not (item.model_id and item.record_id):
        item.record = None
        continue
```

So we can reset the value for the record

lmignon (changes requested):
Nice feature...

connector/checkpoint/checkpoint.py
simahawk: ah, yes, is a computed field but it's better to update the object cache. Thanks.

simahawk: done

lmignon (approved):

PR #219, opened 2016-10-19
Minor improvements and fixes for checkpoint views

simahawk:
1. we don't want to show "Click on..." if there's no record linked!

screenshot from 2016-10-19 16-30-49
1. `message` field was missing in tree view

At some point we should also put the message outside `H1` but this is another story :)

### DIFF
--- a/connector/checkpoint/checkpoint_view.xml
+++ b/connector/checkpoint/checkpoint_view.xml
@@ -21,7 +21,7 @@
                     </header>
                     <sheet>
                         <h1>
-                            <div attrs="{'invisible': [('model_id', '!=', False)]}">
+                            <div attrs="{'invisible': [('model_id', '=', False)]}">
                                 Click on the <field name="model_id" class="oe_inline" options='{"no_open": 1}'/>
                                 to verify it:
                                 <field name="record"/>
--- a/connector/checkpoint/checkpoint_view.xml
+++ b/connector/checkpoint/checkpoint_view.xml
@@ -48,6 +48,7 @@
                         delete="false" edit="false" version="7.0">
                     <field name="model_id"/>
                     <field name="record"/>
+                    <field name="message"/>
                     <field name="backend_id"/>
                     <field name="state"/>
                     <button name="reviewed"


connector/checkpoint/checkpoint_view.xml
```diff
@@ -48,6 +48,7 @@
                         delete="false" edit="false" version="7.0">
                     <field name="model_id"/>
                     <field name="record"/>
+                    <field name="message"/>
```
lasley, line 51: Do we actually want to show this on tree view? It's already pretty filled up as it is IMO

simahawk: well, without this, if you have a message only record you get a line with no meaning at all.
OTOH we could use  `name` and remove `model_id`, `record` and `message` at once.
Maybe we just need to add the description of the model in the name instead of just the display name.
What do you think?

yvaucher (approved):

connector/checkpoint/checkpoint_view.xml
lasley: Ah I see your point. Your latter idea of using `name` seems the best in terms of flexibility IMO.

The downside there would be an inability to sort on those columns, which I'm pretty sure some of my customers do. 

In your use case, how long are these messages? I looked at the default, and it seemed about 50 chars. Are the live messages about as long?

simahawk: well, it depends on what you are tracking. For instance, in the PS connector you can track webservice errors and that could be a long string, but in this case you don't have record and model.
You know what? I don't think it's a real issue with modern screen sizes, so maybe we can just try this way and if in "real life" we recognize is not good we can always find a better solution. Make sense?

lasley: Reason I said something is that my checkpoints screen is already essentially full on a 1200px screen:

image

I won't block though, as I don't have another solution and I can see the use case in your's. These models are poorly named too, due to crazy medical nesting and BS relation tables in the external system.

lasley (approved):

connector/checkpoint/checkpoint_view.xml
pedrobaeza: You need to install a module called _web_responsive_ for saving horizontal space. Do you know it? :laughing:

guewen (approved):

PR #220, opened 2016-10-25
[imp] add checkpoint straight from backend record

simahawk:
Goal: simplify the checkpoint story.

**Now**

ATM you have to define this in your own code

```
from openerp.addons.connector.checkpoint import checkpoint

def add_checkpoint(session, model_name, record_id, backend_id, message=''):
    """Add checkpoint for prestashop backend."""
    return checkpoint.add_checkpoint(session, model_name, record_id,
                                     'prestashop.backend', backend_id)

def add_checkpoint_message(session, backend_id, message=''):
    """Add checkpoint message for prestashop backend."""
    return checkpoint.add_checkpoint_message(
        session, 'prestashop.backend', backend_id, message=message)
```

and then you have to import those functions, etc.

The point is: you always need your backend record to add a checkpoint.

**Improvement**

You just do this

```
backend_record.add_checkpoint(
        model='res.partner', record_id=1, message='VAT number can be missing')
```

or

```
backend_record.add_checkpoint(message='Failed import for category Foo')
```

Tests updated.

### DIFF
--- a/connector/backend_model.py
+++ b/connector/backend_model.py
@@ -48,7 +48,7 @@ class ConnectorBackend(models.AbstractModel):
             raise ValueError('The backend %s has no _backend_type' % self)
         return backend.get_backend(self._backend_type, self.version)
 
-    @api.model
+    @api.multi
     def add_checkpoint(self, model=None, record_id=None, message=''):
         """Add a checkpoint for current backend."""
         self.ensure_one()
--- a/connector/backend_model.py
+++ b/connector/backend_model.py
@@ -49,7 +49,7 @@ class ConnectorBackend(models.AbstractModel):
         return backend.get_backend(self._backend_type, self.version)
 
     @api.multi
-    def add_checkpoint(self, model=None, record_id=None, message=''):
+    def add_checkpoint(self, model=None, record_id=None, message=None):
         """Add a checkpoint for current backend."""
         self.ensure_one()
         cp_model = self.env['connector.checkpoint']
--- a/connector/backend_model.py
+++ b/connector/backend_model.py
@@ -21,7 +21,6 @@
 
 from openerp import models, fields, api
 from . import backend
-from . import checkpoint
 
 
 class ConnectorBackend(models.AbstractModel):
--- a/connector/backend_model.py
+++ b/connector/backend_model.py
@@ -21,6 +21,7 @@
 
 from openerp import models, fields, api
 from . import backend
+from . import checkpoint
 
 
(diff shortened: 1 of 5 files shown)


connector/backend_model.py
```diff
@@ -47,6 +48,20 @@ def get_backend(self):
             raise ValueError('The backend %s has no _backend_type' % self)
         return backend.get_backend(self._backend_type, self.version)
 
+    @api.model
```
lmignon: `@api.multi` otherwise `self.id` can be empty

lmignon (changes requested):

connector/backend_model.py
simahawk: typo fixed, thanks! :)

connector/backend_model.py
```diff
@@ -47,6 +48,20 @@ def get_backend(self):
             raise ValueError('The backend %s has no _backend_type' % self)
         return backend.get_backend(self._backend_type, self.version)
 
+    @api.multi
+    def add_checkpoint(self, model=None, record_id=None, message=''):
```
lmignon: Why a default empty string for param `message`?

connector/checkpoint/checkpoint.py
```diff
@@ -184,6 +184,8 @@ def _needaction_domain_get(self):
         return [('state', '=', 'need_review')]
 
 
+# TODO: deprecate these since we can now add checkpoints
```
lasley: It would be nice if there were a `_logger.warn(` to indicate the deprecation log-side

lasley (changes requested):
Lint is 🔴

simahawk:
@lasley the build is red for docs lint, in OCA repo too https://travis-ci.org/OCA/connector/jobs/170488884
I'll check the reason but I didn't touch all those file, just 2 of them, so I wonder if there's an issue on Travis with such tests.

connector/backend_model.py
simahawk: @lmignon updated

connector/checkpoint/checkpoint.py
simahawk: @lasley done

lmignon (approved):

lmignon:
@simahawk Thank you for these improvements!

simahawk:
@lasley oops, I left behind an import... fixed and build is green again ;)

lasley (approved):
Thanks @simahawk

guewen (approved):

PR #221, opened 2016-11-01
[IMP][8.0] connector: Rename to_openerp by to_odoo method and make alias

sergio-teruel:
@guewen @pedrobaeza @llacroix in connector_prestashop we rename all methods to_openerp by to_odoo, this is a little change as in v9.
Thanks

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -400,7 +400,7 @@ class Binder(ConnectorUnit):
     _openerp_field = 'openerp_id'  # override in sub-classes
     _sync_date_field = 'sync_date'  # override in sub-classes
 
-    def to_openerp(self, external_id, unwrap=False):
+    def to_odoo(self, external_id, unwrap=False):
         """ Give the OpenERP ID for an external ID
 
         :param external_id: external ID for which we want
@@ -421,6 +421,9 @@ class Binder(ConnectorUnit):
         if unwrap:
             bindings = getattr(bindings, self._openerp_field)
         return bindings
+
+    # 2016-10-25: compatible with old implementations
+    to_openerp = to_odoo
 
     def to_backend(self, binding_id, wrap=False):
         """ Give the external ID for an OpenERP binding ID
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -400,7 +400,7 @@ class Binder(ConnectorUnit):
     _openerp_field = 'openerp_id'  # override in sub-classes
     _sync_date_field = 'sync_date'  # override in sub-classes
 
-    def to_odoo(self, external_id, unwrap=False):
+    def to_openerp(self, external_id, unwrap=False):
         """ Give the OpenERP ID for an external ID
 
         :param external_id: external ID for which we want
@@ -422,8 +422,8 @@ class Binder(ConnectorUnit):
             bindings = getattr(bindings, self._openerp_field)
         return bindings
 
-    # 2016-10-25: compatible with old implementations
-    to_openerp = to_odoo
+    def to_odoo(self, external_id, unwrap=False):
+        return self.to_openerp(external_id, unwrap)
 
     def to_backend(self, binding_id, wrap=False):
         """ Give the external ID for an OpenERP binding ID


pedrobaeza (approved):

atchuthan (approved):

connector/connector.py
```diff
@@ -422,6 +422,9 @@ def to_openerp(self, external_id, unwrap=False):
             bindings = getattr(bindings, self._openerp_field)
         return bindings
 
+    # 2016-10-25: compatible with old implementations
+    to_openerp = to_odoo
```
lmignon: is this kind of trick compatible with potential overrides of the `to_openerp` method?

pedrobaeza: No, it isn't, but do you have any override in mind? It can be done the other way (adding the alias `to_odoo`)

lmignon: @pedrobaeza My main concern is to not break the existing API. I've cases where we have override this method to add additional checks and implement a more complex mapping rule. I don't understand the motivation behind this change on V8 and IMO we must avoid this kind of change on a stable release.
:-1:

pedrobaeza: As said, it can be done the other way. @sergio-teruel, please change it and take care about the overrides.

pedrobaeza:
@lmignon, does this look now better?

lmignon:
What's the motivation for this kind of change? Sorry but I don't understand why we need it in 8.0.

pedrobaeza:
We need it for the PrestaShop connector, to make the less changes between 9.0 and 8.0 versions.

lasley:
Both ways are going to introduce fracturing. IMO the original way was actually better, because it will fracture less & in a more backwards compatible manor:

```
>>> class A(object):
...  def to_odoo(self):
...    print 'to_odoo'
...  to_openerp = to_odoo
... 
>>> class B(A):
...   def to_openerp(self):
...     print 'to_openerp'
...     return super(B, self).to_openerp()
...
>>> B().to_openerp()
to_openerp
to_odoo
>>> B().to_odoo()
to_odoo
```

vs:

```
>>> class C(object):
...   def to_openerp(self):
...     print 'to_openerp'
...   def to_odoo(self):
...     print 'to_odoo'
...     return self.to_openerp
... 
>>> class D(C):
...   def to_openerp(self):
...     print 'to_openerp_child'
...     super(D, self).to_openerp()
...   def to_odoo(self):
...     print 'to_odoo_child'
...     super(D, self).to_odoo()
... 
>>> D().to_openerp()
to_openerp_child
to_openerp
>>> D().to_odoo()
to_odoo_child
to_odoo
```

lmignon:
If you need this change for prestashopaddons a better solution is to put it in an extend of the original class in this module. Therefore the scope of this change is limited to the addons for prestashop. IMO it's the better and safer solution.

sergio-teruel:
Ok, we can do that in connector_prestashop module, done in this commit https://github.com/OCA/connector-prestashop/pull/22/commits/fa7bcf857ba43f43e67d89cbed1302b183fecf60#diff-a9b0b1aeac7b3b822176dc95d20ec2d1R43
But in connector module v9 we should do the same. (https://github.com/OCA/connector/blob/9.0/connector/connector.py#L385)

guewen:
We should consider to revert the commit in 9.0 given the incompatibilities raised by @lasley

pedrobaeza:
We can simply drop the alias.

lmignon:
I agree with @guewen we should revert the commit in 9.0

pedrobaeza:
We shouldn't perpetuate the `to_openerp` name. Better to just move on to the name `to_odoo` as the only valid.

lmignon:
`to_odoo` can replace `to_openerp`in V10 not before
I propose to revert the change in 9.0 https://github.com/OCA/connector/pull/224

pedrobaeza:
There's no reason for that delay.

lmignon:
@pedrobaeza and why not in 7.0. We MUST keep a backward compatible api.

pedrobaeza:
Then the same principle would apply to 10.0 as well. From version to version we don't need to keep that compatibility.

guewen:
@pedrobaeza there is no reason to rename before v10, where even odoo itself still uses `openerp` internally. I close the discussion here, this has to be renamed in Odoo 10.0.

llacroix:
@guewen well technically the use of openerp should be deprecated since v8. It's a shame odoo doesn't have a proper deprecation API.

@lmignon v7 was openerp so no need to have backward compatibility here unless you're backporting a module made from odoo v8+. v8+ should obviously be called to_odoo or something else.

Honestly, all of this to_odoo, to_openerp is meaningless. Instead of using to_odoo, or to_openerp it should use a more general term that isn't related to the actual name of the platform that we use... 

Here's what I can see from the pydoc in the connector.py

```
    def to_openerp(self, external_id, unwrap=False):                                                                                        
        """ Give the OpenERP ID for an external ID
        ....

    def to_backend(self, binding_id, wrap=False):
        """ Give the external ID for an OpenERP binding ID 
        ....
```

If we replace OpenERPID by InternalID we have the meaning that one function convert internal id into external id and the other does the inverse.

At first, I believe that the "to_backend" is misleading because in some cases, odoo is usually the actual backend. And the connector backend are actually services... So instead of getting messed with the terminology... lets keep things simple.

```
to_internal_id -> to_odoo/to_openerp
to_external_id -> to_backend
```

That can be backported and ported up to v10+... It won't matter even if odoo gets renamed to Oboo later.. the name is more clear because internal_id quite match the description of the pydoc and `to_external_id` quite match. I have other suggestions such as `get_binding`, `to_external`, `to_internal` (without id, it would return a resultset), `get_external`, `get_inernal`, `to_service`, `to_platform`, `to_local`, `to_remote`.... Choose your pick but please! Dump `to_odoo`, `to_openerp` to the trash and never ever name things like that in odoo because odoo could be renamed to something else... just keep a term that actually means something.

lmignon:
@llacroix my proposal for v7 was a joke...
My main concern is to keep a stable API into each version. I'm open to any changes / refactoring and since connetor 10.0 is not yet released it's the right time to propose changes that breaks the API. Feel free to make a PR to rename these methods in v10.

guewen:
@llacroix Your message seems quite emotional, I suggest that we keep a soft tone. Renaming/deprecating things over 2 versions is not excessive IMO.

I agree that `to_backend` was a poor choice of name because it is confusing with the `backend` and that `to_external` is a better choice.

I mildly disagree that `to_odoo` is a bad choice of name, it's meaningful. You say it should not be related to the platform that we use, which is correct for the service we interact with as it change for each implementation, but for `to_odoo`, we are necessarily using `odoo` so yes, it is meaningful. I might be inclined to rename `to_odoo` if there is a common opinion about that though. But renaming it just because some day Odoo might be renamed is not the most valuable argument, as anyway there would be other things to rename. So if we are to rename it, I propose `to_binding`.

Finally, I strongly disagree to rename anything until Odoo 10.0.

llacroix:
@guewen Sorry if it sound emotional. I'm not being emotional.

About `to_odoo`, I'm not saying it's meaningless, but it's clearly not an improvement over `to_openerp`. There are lots of weird things in odoo/openerp. And we should fix those weird things instead of keeping them to the next version. 

For the `to_binding` naming, It's not that I'm against it... It's just that the name `binding_id` is frequently used in the code and sometimes, it seems that it's the object mapping a remote object and odoo object or sometimes it's the odoo object itself. As I understand, `to_binding` should return the object that bind an odoo object with the remote object. Which isn't what `to_odoo` does. `to_odoo` or `to_openerp` seems to return the actual object bound to the binding. 

> But renaming it just because some day Odoo might be renamed is not the most valuable argument, as anyway there would be other things to rename. So if we are to rename it, I propose to_binding.

I see how that worked out for `to_openerp`.

lasley:
IMO `to_external` and `to_internal` is a more consistent & abstract approach to the naming. `to_binding` doesn't work in all conditions, because you could also be returning the unwrapped, normal record.

I admittedly don't like the current naming, and have them remapped to the to_internal/external in my latest connector.

My two cents.

PR #224, opened 2016-11-09
[FIX] Revert support `to_odoo` to address next v10 release

lmignon:
The rename of `to_openerp` in `to_odoo` is not backwards compatible

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -357,7 +357,7 @@ class Binder(ConnectorUnit):
     _openerp_field = 'openerp_id'  # override in sub-classes
     _sync_date_field = 'sync_date'  # override in sub-classes
 
-    def to_odoo(self, external_id, unwrap=False):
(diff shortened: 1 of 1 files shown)


pedrobaeza:
As stated, I prefer to leave `to_odoo` name and simply remove the alias.

lmignon:
@pedrobaeza as stated, we have code in production that override this method...

guewen:
@pedrobaeza should be done in odoo 10

pedrobaeza:
OK, do what you want.

guewen (approved):

sebalix (approved):

sebalix:
I agree with @lmignon , this breaking change should be done in v10.

pedrobaeza:
OK, merging. @sergio-teruel @simahawk, please take this into account for Prestashop connector for v9.

PR #225, opened 2016-11-13
Allow overriding of requests timeout value for long running tasks 

hardikns:
The current value is set to 1sec. My job does little more than just creating records. When the timeout happens, it sometimes fails to set the job back to pending. I figured that situation will not arise if we don't let the timeout kick in.


sbidoul:
`set_job_pending` is called in all cases. The only reason for the timeout is to log configuration errors which raise after much less than the one sec timeout. 

The root cause of your problems must be something else.

guewen:
Please re-open if needed.

PR #226, opened 2016-11-26
[9.0] improve sequential channels

sbidoul:
This implements a long standing todo regarding the behaviour of sequential channels in presence of ETA.

Jobs that get retried are  now correctly executed in the intended order if they are in a sequential channel.
This is useful for instance when doing long imports with base_import_async where the import order is important.


lmignon (approved):
Code review only

guewen (approved):
Excellent feature, well done. LGTM

PR #228, opened 2016-12-22
Fix E305 pep8 errors previously not detected

guewen:
The error text is:
E305 expected 2 blank lines after class or function definition, found 1


lmignon (approved):
Thank you @guewen

pedrobaeza (approved):

PR #229, opened 2016-12-22
fix(worker): Backport fix for #58 to banch 7.0

agb80:


connector/unit/mapper.py
```diff
@@ -142,7 +142,7 @@ def convert(field, conv_type):
     """
     def modifier(self, record, to_attr):
         value = record[field]
-        if not value:
+        if not value and value != 0:
```
guewen, line 145: there's the same issue with `value == ''` so maybe we should totally change this condition

agb80: What about try with something like this?
`if value is not None:`

PR #231, opened 2017-01-19
[10.0] Reworked components registry

guewen:
At the creation of the connector (OpenERP 7.0), many architecture choices have been made
based on the old API of Odoo. One of my main concern was to get rid of the `cr,
uid, context` in the classes of the connector.  Now, two things have changed:

* Odoo is entirely using its new API
* I have years of look back on what works well or not, plus various
  implementation of connectors in several domains, plus many
  community/colleagues feedbacks.

That brought me to think more and more about the future of the connector and
how to correct the former design mistakes, and at the same time making it
easier to grasp.

The connector is originally composed of:

* A Jobs Queue
* Events (producer, subscriber) pattern
* The Backends registries (`ConnectorUnit`) that I will now call the Components
  registry and a set of base components (`Mapper`, `Binder`, ...)

## Queue Job API

I already started with the Jobs Queue which is now extracted in an [independent
addon](https://github.com/OCA/queue/tree/10.0/queue_job) and has got its API
entirely reviewed. It is now melted into the Odoo API:

```python

class MyModel(models.Model):
   _name = 'my.model'

   @api.multi
   @job
   def my_method(self, a, k=None):
       _logger.info('executed with a: %s and k: %s', a, k)


class MyOtherModel(models.Model):
    _name = 'my.other.model'

    @api.multi
    def button_do_stuff(self):
        self.env['my.model'].with_delay().my_method('a', k=2)

```

## Components API

The basic idea is to keep and improve what works and dump what doesn't (easy
isn't it?), I'll try to make a better component system, more close to the
Odoo API.

What works:

* The component system. It allows to put together the different pieces and
  that's really cool to be able to decouple them.  Being able to "ask" for a class
  without knowing it is powerful.
* The `ConnectorEnvironment` instance which is passed to each new
  `ConnectorUnit` transversally, so we can store info inside and keep it all
  along a synchronization flow.

What is flawed:

* The worst is the extensibility: it's easy to override something in a new
  backend version, but hard to create extension modules to improve an existing
  backend version.
* The backend versions: too hard to use.  Differences between versions of a
  backend reveals to be too rare to justify this feature. Furthermore, this
  is the main reason for the flaw number 1 so let's drop it.
* The lookup of the `ConnectorUnit` by class. It limits some usage as we need
  to know and import the class.  It could be more dynamic and flexible with
  strings.
* Only able to retrieve one component at a time.
* `model_name` was mandatory but in some case we want components without model,
  we had to use fake model names
* If we can reduce the number of imports along the way...

I started a `component` addon, which will be a dependency of the `connector`,
but will also be an independent and agnostic component registry integrated
with Odoo.

The components borrows the `_name` and `_inherit` of Odoo for a good
extensibility, and introduce a `_usage` attribute for the lookup (kind of
informal interface, but components with different `_usage` can share the same
interface). Examples of `_usage` in the know world could be `import.mapper` or
`backend.adapter`.

Components are registered towards a `_collection` which is the name of a Model
acting the same way the `Backend` today and they are applied on models with
`_apply_on`.

So `_name` and `_inherit` are used for extensibility, `_usage` and `_apply_on`
for finding the Components.

When we want to work with the Components, we instanciate a `WorkContext`
(replacement of `ConnectorEnvironment`) with the record of the collection
(backend record), this work context will be passed to every component used
during this work session.

Example:

```python


class ShopBackend(models.Model):

    _name = 'shop.backend'
    _inherit = ['collection.base']

    name = fields.Char()
    url = fields.Char()


class ShopBaseComponent(AbstractComponent):
    _ name = 'shop.base'
    _collection = 'shop.backend'


class SaleImporter(Component):
    _name = 'importer.shop.sale'
    _inherit = ['shop.base', 'importer']  # importer is a base component defined in 'connector'
    _usage = 'importer'  # used for retrieval of the component, in fact, it is already set in the 'importer' Component so we would not need to repeat it here
    _apply_on = ['shop.sale.order']

    def run(self, external_id):
        # ...
        sale_mapper = self.component(
            usage='import.mapper',
            # no need to specify the model we are working with
        )
        partner_mapper = self.component(
            usage='import.mapper',
            model_name='shop.res.partner',  # for a different model
        )
        # ...
        return ...


class SaleImportMapper(Component):
    _name = 'import.mapper.shop.sale'
    _inherit = ['shop.base', 'import.mapper']  # import.mapper is a base component defined in 'connector'

    _usage = 'import.mapper'  # used for retrieval of the component
    _apply_on = ['shop.sale.order']

    def map_record(self, record):
        return ...


class PartnerImportMapper(Component):
    _name = 'import.mapper.shop.partner'
    _inherit = ['shop.base', 'import.mapper']  # 'import.mapper' is a base component defined in 'connector'

    _usage = 'import.mapper'  # used for retrieval of the component
    _apply_on = ['shop.res.partner']

    def map_record(self, record):
        return ...


class SaleOrder(models.Model):

    _inherit = 'shop.sale.order'

    @api.multi
    def import_from_shop(self, external_id):
        backend = self.backend_id
        with backend.work_on('shop.sale.order') as work:
            importer = work.component(usage='importer')
            importer.run(external_id)



```

Also, the new system will allow to search and run all components for a usage.
In the following example, we run a main mapper, then run all mappers with
`_usage` `import.mapper.extras`, meaning that an addon can plug its own mapper
easily (but he might as well use `_inherit` to extend the original one):

```python

  values = self.component(usage='import.mapper').map_record(record).values()
  for mapper in self.many_components(usage='import.mapper.extras'):
      values.update(mapper.map_record(record).values())

```

It seems this design would correct the flaws I mentioned above.

The existing set of base components (`Binder`, `Mapper`, ...) have been
reimplemented in `connector` on top of the Components.

## Events API

Events are now in a new addon `component_event` that depend on `component`. The API got an upgrade too:

```python
# Notification

    class Base(models.AbstractModel):
        _inherit = 'base'

        @api.model
        def create(self, vals):
            record = super(Base, self).create(vals)
            self._event('on_record_create').notify(record, fields=vals.keys())
            return record

# Listeners:

    class MyEventListener(Component):
        _name = 'my.event.listener'
        _inherit = 'base.event.listener'

        def on_record_create(self, record, fields=None):
            _logger.info("%r has been created", record)

```


guewen:
As the PR  #211 for the migration of the connector  is not merged yet, we see too much commits.
What is worth to see is in `component` and perhaps `test_component`.

pedrobaeza:
@sergio-teruel, can you take a look at the proposal?

lmignon:
@guewen Do you reinvent `zope.component` https://pypi.python.org/pypi/zope.component?

guewen:
@lmignon I had a look at zope.component and there's certainly some inspiration. However, my goal is to have something very close to the Odoo API, reusing its extension mechanism (`_inherit`), and much more simple than zope.component.

lmignon:
@guewen You are right to avoid the use of `zope.component`. (As a former user/developer of plone I prefer to to no more use this lib that makes your code unreadable.. IMPOV)

guewen:
This is a very very early work but here is what would change on an implementation (untested, only small parts migrated): 
connector-ecommerce: https://github.com/guewen/connector-ecommerce/commit/82bacb7233b6412a196dca253264dec62fead98b
magento connector: https://github.com/OCA/connector-magento/commit/3986233faffa45ca9a43a943f74c7b90efa1c33e

guewen:
`component` is now fully tested (98%). Still lacking a bit of documentation.

`connector` will (should) be backward compatible, ConnectorUnit or Component classes will both be usable.
I plan to remove the ConnectorUnit / Backend versions features in Odoo 11.0.
I have to update the docs and add tests for the new Components. Some duplication might be removed between ConnectorUnit and Component classes (in mapper notably), however I don't want to spend too much time on that as the former will be dropped.

The Magento Connector implementation is going well, the main module is fully functional, I have to rewrite the tests (using vcr.py) and update the doc. https://github.com/guewen/connector-magento/tree/10.0-components

guewen:
`component` got a round of documentation.

guewen:
[conn-doc.tar.gz](https://github.com/OCA/connector/files/1085402/conn-doc.tar.gz)

Here's a local build of the docs, including a migration guide. /cc @pedrobaeza @sergio-teruel

guewen:
Docs uploaded on http://connector-rework-doc.surge.sh/
Here is the migration guide: http://connector-rework-doc.surge.sh/guides/migration_guide.html#migration-guide

Note: the url is temporary, should be on odoo-connector.com once merged

guewen:
Travis pass, but runbot fails, the instance is up and running though.

guewen:
Unrelated but BTW who set the "Squash and merge" to be the default option for merge? I don't want it to be the default for this project. Can I have back my access to the admin settings on this project? (I can't even change the default branch and so on)

lmignon:
> Unrelated but BTW who set the "Squash and merge" to be the default option for merge?

@guewen For sure it's not me :smirk:

pedrobaeza:
@guewen AFAIK, this is recognized from your last choice merging something. It's not a project setting, so don't blame me :stuck_out_tongue_winking_eye:

guewen:
> @guewen AFAIK, this is recognized from your last choice merging something. It's not a project setting, so don't blame me

I'm not! 

Yes you right, the only setting is if we want to allow the merge/squash/rebase or not. I would still be glad to have more rights on the project :) But at least someone setting the 10.0 branch as default :)

sbidoul:
@guewen I gave Admin access to the Connectors Maintainer teams (@lmignon, @lasley and yourself so that should be safe). 

Fantastic work here, btw!

guewen:
@sbidoul thanks!

component_event/models/base.py
```diff
@@ -0,0 +1,112 @@
...
+class Base(models.AbstractModel):
+    """ The base model, which is implicitly inherited by all models.
+
+    Add an :meth:`_event` method to all Models. This method allows to
+    trigger events.
+
+    It also notifies the following events:
+
+    * ``on_record_create``
+    * ``on_record_multi``
```
lmignon: @guewen s/on_record_multi/on_record_write

lmignon:
@guewen Your PR is amazing! It's a long reading to understand all the changes and the benefits of this great refactoring. I'll try to use this new version in the next weeks on my current project to validate this new architecture. Thank you for this great work!

guewen:
@here if anyone is currently trying this branch, you should be aware of the change of a8e1778:

> Change Collection.work_on() to a context manager
It allows to open/close objects on start/end of a synchro / work
session. The base method does none of that, but this is a really common
use case. Example: at the beginning of the work session, I open a
webservice client that is available in our WorkContext for the duration
of our work. At the end, I close the client to end the connection.

Example in magento connector:

```
    @contextmanager
    @api.multi
    def work_on(self, model_name, **kwargs):
        self.ensure_one()
        lang = self.default_lang_id
        if lang.code != self.env.context.get('lang'):
            self = self.with_context(lang=lang.code)
        magento_location = MagentoLocation(
            self.location,
            self.username,
            self.password,
            use_custom_api_path=self.use_custom_api_path
        )
        # We create a Magento Client API here, so we can create the
        # client once (lazily on the first use) and propagate it
        # through all the sync session, instead of recreating a client
        # in each backend adapter usage.
        with MagentoAPI(magento_location) as magento_api:
            _super = super(MagentoBackend, self)
            # from the components we'll be able to do: self.work.magento_api
            with _super.work_on(
                    model_name, magento_api=magento_api, **kwargs) as work:
                yield work


```

lmignon:
@guewen What's the maturity of this code? Is-it safe to base new devs on this code?

guewen:
@lmignon the magento connector implementation is working with this code (still working on it but the Components part is done) and we are working on a very custom connector based on it. There is a small risk that I need to change a public API if we hit an issue, but nothing too large I presume, it'd more be adjustments (like `Collection.work_on()` that became a context manager). So yes, I encourage to base new devs on this code, which also gives you the opportunity to propose API fine-tunings before it is frozen.

lmignon:
@guewen Thank you for your response. You did an awesome job!  I migrated the code from a very specific connector with many components in less than half a day. The API is clean and the code is much cleaner!  The only difficulty encountered was to run the tests, which gave me the opportunity to analyze the code more in depth to fix the problem. (My first small contribution to this big work :smirk:). I agree with you to encourage to base new devs on this code.

guewen:
@lmignon happy to read that! And thanks for your contribution, I pushed a new commit that benefits from it already, I could simplify the existing tests and add a ConnectorTransactionCase that loads all the base components in the test `self.comp_registry`.

connector/tests/common.py
```diff
@@ -1,47 +1,63 @@
...
-        kwargs.pop('description', None)
-        job_func(*args, **kwargs)
-
-    with mock.patch(job_path) as patched_job:
-        # call the function directly instead of '.delay()'
-        patched_job.delay.side_effect = clean_args_for_func
-        yield patched_job
+
+    def setUp(self):
+        super(ConnectorTransactionCase, self).setUp()
```
lmignon: @guewen `setUpClass` is more appropriate to avoid a reload of the registry for each test.
IMO this changes breaks the modularity. If you write a test for your components you need to know all the modules you depend on that provide components to manually load their components.
You can safely rely on the `_register_hook` method of the `component.builder`model to load all the components of the modules already installed since the method will be called at the end of the registry loading process AND the method clear/clean the registry at each call. Therefore you must only call the `load_components` method for the addon you are testing. Do I miss something?

guewen: @lmignon but `_register_hook` register components in the global component registry for the database, here I want them in the test `comp_registry`. I extracted the content of `_register_hook` in a `build_registry` so we can use it from the setup: 834b351

Indeed, it works well and is more extensible!

Regarding `setUpClass`,  we need to reset the test component registry between each test, not for a whole class. The reason is that if inside a test you add a new component, it should no longer be in the registry for the next test.

```python
    def test_register(self):
        """ Able to register components in components registry """
        class Component1(Component):
            _name = 'component1'

        class Component2(Component):
            _name = 'component2'

        # build the 'final classes' for the components and check that we find
        # them in the components registry
        Component1._build_component(self.comp_registry)
        Component2._build_component(self.comp_registry)
        self.assertEquals(
            ['base', 'component1', 'component2'],
            list(self.comp_registry)
        )

    def test_inherit_bases(self):
        """ Check __bases__ of Component with _inherit """
        class Component1(Component):
            _name = 'component1'

        class Component2(Component):
            _inherit = 'component1'

        class Component3(Component):
            _inherit = 'component1'

        Component1._build_component(self.comp_registry)
        Component2._build_component(self.comp_registry)
        Component3._build_component(self.comp_registry)
        self.assertEquals(
            (Component3,
             Component2,
             Component1,
             self.comp_registry['base']),
            self.comp_registry['component1'].__bases__
        )
```

Cedric-Pigeon:
@guewen Fantastic job I based my Lengow connector on your work and I really enjoy the new api. Thanks for that.

guewen:
@lmignon could you have a look at 1f448f6? Is it more or less what you had in head?

component/builder.py
```diff
@@ -49,13 +49,20 @@ def _register_hook(self):
...
             cachesize=self._components_registry_cache_size
         )
         _component_databases[self.env.cr.dbname] = components_registry
-        self.build_registry(components_registry)
+        return components_registry
 
-    def build_registry(self, components_registry):
+    def build_registry(self, components_registry, states=None):
+        if not states:
+            states = ('installed', 'to upgrade')
```
guewen: I tend to think we don't even need to load the 'to upgrade' components at all?

guewen:
IMO this branch is now stable enough to be merged :fire: 

Not sure why runbot is red, the log shows 'success', and I'm able to connect on the runbot instance.

We are using it for the Magento connector BTW:
* https://github.com/OCA/connector-ecommerce/pull/41
* https://github.com/OCA/connector-magento/pull/255

lmignon:
@guewen sorry for the delay. Back from holidays I'll review your last changes ASAP...

guewen:
> sorry for the delay. Back from holidays I'll review your last changes ASAP...

Really no need to be sorry, holidays well deserved :)

jaredkipe:
Inside `core.MetaComponent._build_component` the documentation comments has a hierarchy with two classes `B1` and `B2` that both define `_name = 'b'` but that is not allowed by the check on line 765 that checks if the `_name` is already in the registry.

guewen:
> Inside core.MetaComponent._build_component the documentation comments has a hierarchy with two classes B1 and B2 that both define _name = 'b' but that is not allowed by the check on line 765 that checks if the _name is already in the registry.

Good catch! I'm on holidays, will change that when I'm back.

test_component/__manifest__.py
```diff
@@ -0,0 +1,17 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+{'name': 'Components Tests',
+ 'summary': 'Automated tests for Components, do not install.',
```
flotho, line 6: Hi @guewen ,

Could you add a description tag so that Travis pass all the build ?

guewen: Done, Travis already passes the build though. Runbot doesn't for an unknown reason but runbot should not block a merge (the runbot instance is up and running BTW).

guewen: Actually this is the reason why runbot was unhappy :) thanks for the hint

guewen:
@lmignon @lasley @jaredkipe @flotho 
What do you think about merging? Let me know and I'll run a rebase (don't merge before the rebase because I the history has fixup commits).

Once merged, I'll increase the version number, and change the documentation build to use the 10.0 by default (and if time permits, I'll build and push 9.0 and 8.0 docs in subdirectories and add links between versions).

lmignon (approved):
Great Job!

Cedric-Pigeon (approved):
Good for me. Great job ! Thanks.

simahawk:
:+1:

jaredkipe:
Works great for me!

lasley (approved):
Absolutely amazing @guewen

lmignon:
:champagne:

yucer:
Hello @guewen,

Which **OCA** connector project do you keep updated to the last version of the connector framework ?

guewen:
I can't keep one project updated as it depends on the projects we need for our customers.
Anyway, from the top of my head, the version 10.0 of connector-magento is using the components: https://github.com/OCA/connector-magento/tree/10.0

https://github.com/OCA/connector-jira is as well and is probably the last one I worked on.

PR #232, opened 2017-01-20
[9.0][FIX] connector: Show connector tab in partner view only for con nector manager security group.

sergio-teruel:
Minor fix.
If we activate the portal for external user this tab is showed.

cc @Tecnativa @guewen

### DIFF
--- a/connector/res_partner_view.xml
+++ b/connector/res_partner_view.xml
@@ -1,20 +1,18 @@
 <?xml version="1.0" encoding="utf-8"?>
-<odoo>
-    <data noupdate="0">
-        <record id="view_partner_connector_form" model="ir.ui.view">
-            <field name="name">res.partner.connector.form</field>
-            <field name="model">res.partner</field>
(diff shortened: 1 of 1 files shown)


pedrobaeza (changes requested):
Put the group directly for the view.

sergio-teruel:
@pedrobaeza Better of course. Thanks!! :smile:

pedrobaeza (approved):

sebalix (approved):

connector/res_partner_view.xml
```diff
@@ -1,10 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
-<openerp>
+<odoo>
     <data noupdate="0">
```
carlosdauden: You can use `<odoo noupdate="0">` directly instead of add `<data>`.
Although in this case it is not necessary to use noupdate because 0 is the default value.

carlosdauden (approved):
Minor suggestions

PR #235, opened 2017-03-06
[9.0][IMP]connector: multi action for button done on queu job

zakiuu:
Allow multiple job selection to mark them as done

### DIFF
--- a/connector/queue/model_view.xml
+++ b/connector/queue/model_view.xml
@@ -140,6 +140,21 @@
             <field name="search_view_id" ref="view_queue_job_search"/>
         </record>
 
+        <record id="action_button_done_queue_job" model="ir.actions.server">
+            <field name="name">Set to 'Done'</field>
+            <field name="model_id" ref="model_queue_job"/>
+            <field name="state">code</field>
+            <field name="code"><![CDATA[
+self.button_done(cr, uid, context['active_ids'], context=context)
+]]></field>
+        </record>
+
+        <record model="ir.values" id="button_done_queue_job">
+            <field name="name">Set to 'Done'</field>
+            <field name="key2">client_action_multi</field>
+            <field name="value" eval="'ir.actions.server,%d'%action_button_done_queue_job"/>
+            <field name="model">queue.job</field>
+        </record>
 
         <record id="view_requeue_job" model="ir.ui.view">
             <field name="name">Requeue Jobs</field>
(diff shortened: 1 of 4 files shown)


zakiuu:
@lmignon @sbidoul

ThomasBinsfeld:
@zakiuu Works well (functional tests). But IMO, the action button shouldn't overwrite the `date_done` value if the job is already done.

connector/queue/model_view.xml
```diff
@@ -140,6 +140,21 @@
             <field name="search_view_id" ref="view_queue_job_search"/>
         </record>
 
+        <record id="action_button_done_queue_job" model="ir.actions.server">
+            <field name="name">Set to 'Done'</field>
+            <field name="model_id" ref="model_queue_job"/>
+            <field name="state">code</field>
+            <field name="code"><![CDATA[
+self.button_done(cr, uid, context['active_ids'], context=context)
```
lmignon, line 148: @zakiuu uyou should modify the `button_done` method https://github.com/acsone/connector/blob/4d21e60e250a5bd9111b2808c24f7610264251cb/connector/queue/model.py#L133 to filter out jobs not into the expected state.
```python
 @api.multi
    def button_done(self):
        result = _('Manually set to done by %s') % self.env.user.name
        self.filtered(lambda j: j.state in (PENDING, ENQUEUED, FAILED)._change_job_state(DONE, result=result)
        return True
```

lmignon (changes requested):
@zakiuu Thank you for contribution. Can you improve a little bit your PR PLZ.

zakiuu:
@lmignon can you please check 5c836f9

connector/queue/model_view.xml
lmignon: @zakiuu A test into the testsuite would be greatly appreciated.
* set the job state to done,
*  mock the _change_job_sate method (`with mock.patch(Job, '_change_job_state', .... as mocked_change_job:` )
*  call job.button_done()
*  check that the mock has not been called. (`self.assertEqual(0, mocked_change_job.call_count)`)

lmignon: @zakiuu Ok I just saw your test.

lmignon (approved):

ThomasBinsfeld (approved):

PR #245, opened 2017-05-16
[IMP] Add possibility to ignore backend

OpenCode:
An external software we use to connect to Odoo, create a new database every year. Every copy of database has the some information inside. If I create 2 backend for every database, the information (at first run of import) are duplicated because connector search information separated for backend id.

With this parameter, it's possibile to ignore backend.

### DIFF
--- a/connector/connector.py
+++ b/connector/connector.py
@@ -401,21 +401,25 @@ class Binder(ConnectorUnit):
     _openerp_field = 'openerp_id'  # override in sub-classes
     _sync_date_field = 'sync_date'  # override in sub-classes
 
-    def to_openerp(self, external_id, unwrap=False):
+    def to_openerp(self, external_id, unwrap=False, ignore_backend=False):
         """ Give the OpenERP ID for an external ID
 
         :param external_id: external ID for which we want
                             the OpenERP ID
         :param unwrap: if True, returns the normal record
                        else return the binding record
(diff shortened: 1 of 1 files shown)


guewen:
Thanks for the proposal. It seems rather specific to your implementation, and I would suggest you to implement your own Binder in your connector.

OpenCode:
I propose PR because I thing this implementation can be useful for others. In my situation I use functions of connector and I can't bypass this problem with my binder

guewen:
> I propose PR because I thing this implementation can be useful for others. In my situation I use functions of connector and I can't bypass this problem with my binder

Why? I see no reason you couldn't do. Can you point me to some code so I can see what is blocking you?

flotho:
Hi,
IMHO, working with a state on backend could be better. 
As an example, one of our customer changed its Prestashop recently and we had to face 2 simultaneous PS backend with huge quantity of job failed because we set an absurd location for preventing double synchro. We really would like to be able to deactivate backends and open new without problem.

@OpenCode don't you think active / state on Backend could solve your issue?

guewen:
@flotho  I used a state in a custom implementation too. In the end, this should be handled in the implementations.

OpenCode:
@flotho Set a state is not a solution for my situation because we have 2 (or more) backend active in the some moment

PR #247, opened 2017-05-24
[ADD] configurable scheme, host and http auth

nilshamerlinck:
In some deployments we can have multiple Odoo instances running on the same host, each of them listening on a different interface.

This PR makes it possible to configure the host part of the URL called by the Job Runner to run jobs, instead of the currently harcoded "localhost".

Two alternative methods to configure said host:
- standard Odoo _xmlrpc_interface_ parameter
- _ODOO_CONNECTOR_HOST_ environment variable

### DIFF
--- a/connector/jobrunner/__init__.py
+++ b/connector/jobrunner/__init__.py
@@ -48,8 +48,25 @@ class ConnectorRunnerThread(Thread):
     def __init__(self):
         Thread.__init__(self)
         self.daemon = True
-        port = os.environ.get('ODOO_CONNECTOR_PORT') or config['xmlrpc_port']
-        self.runner = ConnectorRunner(port or 8069)
+        scheme = (os.environ.get('ODOO_CONNECTOR_SCHEME') or
+                  config.misc.get("options-connector", {}).get('scheme'))
+        host = (os.environ.get('ODOO_CONNECTOR_HOST') or
+                config.misc.get("options-connector", {}).get('host') or
+                config['xmlrpc_interface'])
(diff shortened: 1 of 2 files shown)


sbidoul:
Looks good to me.

I suppose the backward incompatible API change is not an issue. 

Can you bump the minor version number, though?

guewen (approved):
Great! Thanks

nilshamerlinck:
@guewen @sbidoul I went a step further to allow more complex deployments. For example:
- one Odoo instance with the Job Runner
- a load balancer (haproxy) configured with HTTP Auth and SSL
- multiple Odoo instances behind, running the jobs

sbidoul:
Hmm, needs rebase due to the version number. Actually, bumping the version should be done by the maintainer who merges.

Also could you update the documentation?

I'm curious about the use case. How many workers do you have across all these instances?

nilshamerlinck:
@sbidoul I will update the documentation once https://github.com/OCA/connector/pull/248 will have been merged, to avoid conflicts.

About the use case, it's a big deployment with lots of jobs to process, so we need them to be load balanced as well :)

guewen:
@nilshamerlinck #248  is merged

lasley (changes requested):
Excellent feature @nilshamerlinck. 

#248 is merged, so this is good once rebased and the docs updated as mentioned in https://github.com/OCA/connector/pull/247#issuecomment-307622551

nilshamerlinck:
Done @lasley, sorry for the delay.

guewen (approved):
Thanks!

lmignon (approved):

PR #248, opened 2017-05-24
[ADD] configurable db host and port

nilshamerlinck:
In some deployments, we might want the Job Runner to connect to the database through a different host/port couple than Odoo.

This is required for example if you have a PgBouncer configured in transaction mode: the Job Runner relies on LISTEN/NOTIFY features, which are not supported by said transaction mode. This PR makes it possible to bypass PgBouncer by adding support for _jobrunner_db_host_ and _jobrunner_db_port_ parameters in Odoo's configuration file.

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -43,24 +43,25 @@ How does it work?
 How to use it?
 --------------
 
-* Optionally adjust your configuration through environment variables:
+* By default, the job runner will use Odoo's configuration:
 
-  - set ``ODOO_CONNECTOR_CHANNELS=root:4`` (or any other channels
-    configuration) if you don't want the default ``root:1``.
+  - connect to Odoo via port ``xmlrpc_port``, or ``8069`` if unset
+  - connect to the database via ``db_host`` and ``db_port``
 
-  - if ``xmlrpc-port`` is not set, you can set it for the jobrunner only with:
-    ``ODOO_CONNECTOR_PORT=8069``.
+* To adjust these values, you can either use environment variables:
 
-* Alternatively, configure the channels through the Odoo configuration
-  file, like:
+  - ``ODOO_CONNECTOR_PORT=9069``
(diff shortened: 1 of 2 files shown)


connector/jobrunner/runner.py
```diff
@@ -160,12 +160,25 @@ def _openerp_now():
     dt = datetime.datetime.utcnow()
     return _datetime_to_epoch(dt)
 
+def _connection_info_for(db_name):
+    connection_info = {'database': db_name}
+    for p in ('host', 'port', 'user', 'password'):
+        cfg = config['db_' + p]
```
simahawk: this is going to fail when the key is not there, isn't it?
```
>>> from openerp.tools import config
>>> config['foo']
  File "<console>", line 1, in <module>
  File "/opt/odoo/src/openerp/tools/config.py", line 555, in __getitem__
    return self.options[key]
KeyError: 'foo'
```

simahawk (changes requested):

connector/jobrunner/runner.py
nilshamerlinck: There will always be values for these `db_*` keys, because of the default values defined in `config` initialization.

Actually, I re-used the same code as https://github.com/odoo/odoo/blob/9.0/openerp/sql_db.py#L673

guewen: Why not fallback to `openerp.sql_db.connection_info_for` when there is no key `job_runner_db_host`, so we would benefits from the uri parsing and have less duplication?

sbidoul:
Sounds good, could you update the documentation?

nilshamerlinck:
@sbidoul I removed the part about channels in `runner.py` as it's covered later in `jobrunner.rst`. Note that I intend to work again on the documentation for https://github.com/OCA/connector/pull/247

sbidoul:
@nilshamerlinck a few flake8 errors, otherwise looks good.

guewen (approved):

PR #249, opened 2017-05-25
[add] jobs filter `Created last 24h`

simahawk:
When you check the queue most of the times
you want to see jobs created today
while you are "playing" with the connector.

I can add this to `queue_job` on v10 if consireded useful.

### DIFF
--- a/connector/__openerp__.py
+++ b/connector/__openerp__.py
@@ -20,7 +20,7 @@
 ##############################################################################
 
 {'name': 'Connector',
- 'version': '9.0.1.0.3',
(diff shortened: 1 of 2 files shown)


guewen (approved):
Nice

pedrobaeza (approved):

pedrobaeza:
But please change version number

lmignon (approved):

simahawk:
version updated

sbidoul:
It is considered useful :wink: Thanks!

PR #250, opened 2017-05-29
10.0 mig connector job subscribe mgs

GillesTephaneMeyomesse:


guewen:
Shouldn't it be renamed to `queue_job_subscribe` and be moved in https://github.com/OCA/queue?

sbidoul:
@guewen indeed.

PR #253, opened 2017-06-06
fix(mapper): mapp zero values as numeric instead of False

### LINKED ISSUE #252: Problem exporting integer field with zero value
For the Drupal connector we have being working for several time now we have the situation that we need to export the product.category sequence field to Drupal float field. 

By default odoo set sequence to zero (0) and that's ok. The problem is that when trying to export the field connector is replacing zero (0) value with False that translated to json is becoming an empty string. That empty string is causing Drupal to fail with database error: Incorrect integer value: '' for column 'weight'.

The cause for this change is condition `if not value` on line https://github.com/OCA/connector/blob/8.0/connector/unit/mapper.py#L146 as on python zero is equal to False.

PD: If any is interested you can find Drupal connector here: https://github.com/OpenPymeMx/connector-drupal

agb80:
Zero value is not Null neither False. This commit rewrittes the validation on mapper to ensure False

is only sended to remote service when Null value is on odoo field.

Fix #252

### DIFF
--- a/connector/unit/mapper.py
+++ b/connector/unit/mapper.py
@@ -143,7 +143,7 @@ def convert(field, conv_type):
     """
     def modifier(self, record, to_attr):
         value = record[field]
-        if not value:
+        if value is None:
             return False
         return conv_type(value)
     return modifier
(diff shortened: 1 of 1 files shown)


guewen (approved):
I agree but this might change the current behavior of existing implementations, not sure about what we should do then.

simahawk (approved):
@guewen we could log a warning at startup maybe? Or just add a big comment in README for next release... :)

agb80:
I guess the comment in README will be the best way to go as a warning at startup it's easy to miss.

lmignon (approved):

PR #254, opened 2017-06-08
[FIX] mapper: Remove orphan method _map_children

lmignon:
This method is not used by the connector and generate errors with pylint if it's not implemented in concrete class.

CC @guewen


guewen (approved):

lmignon:
@guewen I've changed the target branch for this PR since the fix is for 10.0 :smirk:

rousseldenis (approved):
:+1:

PR #256, opened 2017-08-18
[ADD] [WIP] Migration script for connector module

cubells:
cc @Tecnativa

### DIFF
--- a/connector/migrations/10.0.1.0.0/pre-migration.py
+++ b/connector/migrations/10.0.1.0.0/pre-migration.py
@@ -1,0 +1,13 @@
+# -*- coding: utf-8 -*-
+# Copyright 2017 Tecnativa - Vicent Cubells
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html).
+
+
+def migrate(cr, version):
+    if not version:
+        return
+    # Delete noupdate data
+    cr.execute("""
+        DELETE FROM ir_model_data
+        WHERE name = 'queue_job_comp_rule'
+    """)


guewen:
Could you explain what queue_job_comp_rule is?

cubells:
this:

https://github.com/OCA/connector/blob/9.0/connector/security/connector_security.xml#L20

cubells:
Maybe it's not needed because now is here:

https://github.com/OCA/queue/blob/10.0/queue_job/security/security.xml#L20

What do you think?

guewen:
could you filter on the module as well? `module = 'connector'`

guewen:
Is it gonna work for migration between 9.0 and 10.0?

cubells:
Yes, it's for migration between 9.0 and 10.0.

guewen:
The group should be removed as well, as you say, it has been moved in queue

pedrobaeza:
@cubells you haven't got the full picture of a migration of this module from v9 to v10. Let me explain: in v10, the module has been unfolded in 2: `queue_job` in OCA/queue, and `connector` here, which depends on the first.

As there's no way to know which features do you need for your migrated v9 DB, the strategy should be:

* Rename on OpenUpgrade `connector` to `queue_job` in apriori.py.
* Include a migration script on v10 `queue_job` that looks for module `connector` and change its state to `to upgrade` (this is the real value to put).
* You have to analyze changes from v9 to v10, including noupdate deleted records and so on, and put it on the proper module (queue_job or connector), but in this case I see that you are looking in v10 queue_job module, which I don't see correct.

cubells:
ok. I redo this pr after analyze all.

cubells:
Related: https://github.com/OCA/OpenUpgrade/pull/1038

cubells:
Related: https://github.com/OCA/queue/pull/20

cubells:
@guewen 
this pr is not necessary because all noupdate xml-ids are moved to module connector or queue_job in v. 10.0

With these prs
https://github.com/OCA/OpenUpgrade/pull/1038
https://github.com/OCA/queue/pull/20
all is done.

I close this pr.

PR #259, opened 2017-09-06
[ADD] 8.0 Take into account HTTP response codes when assigning work

lembregtse:
In our proxy setup we've foreseen potential cases where the proxy would answer a jobrunner with a 503 HTTP error code in case the chosen Odoo backend server would be too busy for answering the jobrunners call while assigning a job.

We would like to propose this change to take into account HTTP error response codes (400 through 600) and just raise when this happens. This will cause the same behaviour as when having a timeout on our call.

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -175,7 +175,12 @@ def _async_http_get(port, db_name, job_uuid):
         try:
             # we are not interested in the result, so we set a short timeout
             # but not too short so we trap and log hard configuration errors
-            requests.get(url, timeout=1)
+            response = requests.get(url, timeout=1)
+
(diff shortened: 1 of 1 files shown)


guewen (approved):
I observed few times jobs being stuck in `enqueued` but could not analyze further. I suspect it might come from this. Thanks a lot!

lmignon (approved):
Thank you @lembregtse Indeed it makes sens to take into account the http status into the response.

simahawk (approved):

guewen:
Thanks again @lembregtse 
I port forwarded your commit for 9.0 in https://github.com/OCA/connector/pull/261 and 10.0 in https://github.com/OCA/queue/pull/25
Review should be trivial as the code is the same :)

PR #260, opened 2017-09-07
[ADD] filter for children values after mapping

yucer:
The filter was previously applied before executing the mapping for the
item values, via the `skip_item` hook.

In the cases that the filter depends of the results of the mapping the
additional hook `skip_add_item` is useful.

It allows to skip the action adding of the item to the result list.

Also the unit tests for both filters have been added.

### DIFF
--- a/connector/tests/test_mapper.py
+++ b/connector/tests/test_mapper.py
@@ -697,6 +697,69 @@ class test_mapper_binding(common.TransactionCase):
                     }
         self.assertEqual(map_record.values(for_create=True), expected)
 
+    def test_mapping_record_children_filters(self):
+        """ Map a record with children, using defined MapChild """
+
+        backend = Backend('backend', '42')
+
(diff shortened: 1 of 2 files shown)


guewen (approved):
This use case makes sense. Thanks for the detailed commit, docstring and the test!

Travis complains about a tiny detail to resolve before merging:
connector/tests/test_mapper.py:752:18: E124 closing bracket does not match visual indentation

yucer:
I did try again in a new pull request.

It is curious, I have pylint with the [python](https://code.visualstudio.com/docs/languages/python)  extension from vscode and it tells me that is OK.

Is there any doc or something to setup the **OCA** linter settings with **vscode** ?

lasley:
Not sure about vscode, but it sounds like you still need `pylint-odoo`. https://github.com/OCA/pylint-odoo#install

PR #262, opened 2017-09-08
[ADD] filter for children values after mapping

yucer:
The filter was previously applied before executing the mapping for the
item values, via the `skip_item` hook.

In the cases that the filter depends of the results of the mapping the
additional hook `skip_add_item` is useful.

It allows to skip the action adding of the item to the result list.

Also the unit tests for both filters have been added.

### DIFF
--- a/connector/tests/test_mapper.py
+++ b/connector/tests/test_mapper.py
@@ -697,6 +697,69 @@ class test_mapper_binding(common.TransactionCase):
                     }
         self.assertEqual(map_record.values(for_create=True), expected)
 
+    def test_mapping_record_children_filters(self):
+        """ Map a record with children, using defined MapChild """
+
+        backend = Backend('backend', '42')
+
+        @backend
+        class LineMapper(ImportMapper):
+            _model_name = 'res.currency.rate'
+            direct = [('name', 'name')]
+
+            @mapping
+            def price(self, record):
+                return {'rate': record['rate'] * 2}
+
+        @backend
+        class SaleLineImportMapChild(ImportMapChild):
+            _model_name = 'res.currency.rate'
+
+            def format_items(self, items_values):
+                return [('ABC', values) for values in items_values]
+
+            def skip_item(self, map_record):
+                '''Filter by the values before the mapping'''
+                record = map_record.source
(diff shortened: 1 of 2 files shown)


rousseldenis:
@yucer @guewen I've an other issue at exactly this line of code.

If get_item_values() function return void list, it returns [(0, 0, {})].

I will make a PR (I'm on 10.0).

yucer:
@guewen I did submit the pull request again the v8 branch. But this is supposed to work also in other branches.

How do you proceed in those cases ? Should I make pull request for each branch ? Does the one who merge perform this task ?

yucer:
@rousseldenis which line do you mean ?

Does it happen with the modifications on this pull request ? Or was it happening before ?

rousseldenis:
@yucer Nope, this happened before but that concern the value in item_values

yucer:
What is still missing here ? 

What can I do to help merge this pull request ?

Am I missing something from the workflow ?

rousseldenis:
@yucer Can you close the PR you don't want to merge as this is a duplicate of #260 ?

After that, this can be a fast merge as it is a simple hook.

yucer:
I did close the other one. What about now ?

Who will merge it ?

Should I submit a PR for very version ? Or just this one ?

rousseldenis:
@guewen @lmignon This can be a fast merge as it is a simple hook.

@yucer Is it planned to be forwardported ?

rousseldenis (approved):

lmignon (approved):

yucer:
@rousseldenis : I guess yes, otherwise there would be a new feature only in an old version.

The reason why I'd started with v8 is because I did need that scenario first in v8. 

But I guess the code match the new versions right ? It would serve me to get the idea of how do you contribute to multiple branches.

Should the contributor submit a PR for each branch?

lasley (approved):

lasley:
> Should the contributor submit a PR for each branch?

Yeah that would be amazing @yucer. You should be able to just cherry-pick the commit into new branches based off of other versions, possibly resolving a conflict or two.

PR #267, opened 2017-10-09
[10] Generic exporter extracted from Magento connector

sebastienbeau:
Hi @guewen 
I am working an shopinvader project and I need to export some record into locomotivecms (mostly customer).
In 8 version have a done a "connector_generic" module with the generic stuff (mostly copy paste from magento project, and adding some helper for the consommer).
In 10 connector project really rock ! I really really like the new component solution, really it just what I need ! So now the only generic feature that exist in my module is the exporter (which was a copy/paste from magento).
In order to avoid copy/pasting code I think we can add a generic export based on the magento exporter.
This PR is a first version of the extraction, and it's work for shopinvader and magento.

What do you think of the idea?

### DIFF
--- a/connector/components/synchronizer.py
+++ b/connector/components/synchronizer.py
@@ -14,8 +14,17 @@ the backend adapters to read or write data on the backend and the binders to
 create the link between them.
 
 """
+import logging
+import psycopg2
 
 from odoo.addons.component.core import AbstractComponent
+from contextlib import contextmanager
+from odoo.addons.connector.exception import (IDMissingInBackend,
+                                             RetryableJobError)
+from odoo import _
+import odoo
+
+_logger = logging.getLogger(__name__)
 
 
 class Synchronizer(AbstractComponent):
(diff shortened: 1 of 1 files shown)


simahawk:
@sebastienbeau hey :) Guewen is on holiday but we already discussed about that and indeed I was planning to do the same. We have lots of reusable bits all around in community and customers modules that we want to put here as a base. I'll try to review your work in the next days.

lmignon:
@sebastienbeau IMO, this kind of base component should go into a dedicated addon (generic_component, base_component, ????). IMO, the component addon should stay agnostic of its potential usage. Nevertheless, for sure it's important to find a way to share the reusable parts of existing modules.

simahawk:
@lmignon then we agree: actually I forgot to mention that we were thinking about some extra module like that... let's pick a proper name... base_component or [connector|component]_mixin (I vote for this :wink:) or whatever

sebastienbeau:
Yes indeed putting this base component in a separated addons is a better idea ! Let's wait @guewen from comming back

guewen:
> IMO, this kind of base component should go into a dedicated addon (generic_component, base_component, ????).

Yes, dedicated addon == connector?

As we now have the `component` addon, I have no problem to have a more opinionated 'connector' addon. I'm not sure I see any good reason not to have this into the connector addon in fact.

sebastienbeau:
I am totaly ok to put it directly into "connector" module. Indeed it's base feature so we can have it by default in "connector"

lmignon (approved):

guewen (approved):

PR #268, opened 2017-10-27
[FIX] component: Cachetools versioning

lasley:
* Change cachetools version to >=2.0.1 in requirements and add to setup to fix OCA#263

### DIFF
--- a/setup/component/setup.py
+++ b/setup/component/setup.py
@@ -1,7 +1,13 @@
 import setuptools
 
 setuptools.setup(
-    install_requires=['cachetools>=2.0.1'],
     setup_requires=['setuptools-odoo'],
     odoo_addon=True,
+    odoo_addons={
+        'external_dependencies_override': {
+            'python': {
+                'cachetools': 'cachetools>2.0.1',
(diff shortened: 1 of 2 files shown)


simahawk (approved):

lmignon (approved):

setup/component/setup.py
```diff
@@ -1,6 +1,7 @@
 import setuptools
 
 setuptools.setup(
+    install_requires=['cachetools>=2.0.1'],
     setup_requires=['setuptools-odoo'],
     odoo_addon=True,
```
sbidoul: Since that dependency is in `__manifest__.py` it's better to do this (otherwise you may end up with duplicates in the generated install_requires.

```python
odoo_addons={
    'external_dependencies_override': {
        'python': {
            'cachetools': 'cachetools>2.0.1',
        }
    }
}
```

lasley: Makes sense there's a mechanism for that. This what we're looking for? 5f778e0

guewen:
Thanks!

PR #270, opened 2017-12-22
Use a requests session to limit opened sessions

guewen:
We have some issues with a very high throughput of jobs and Odoo's
anonymous sessions. Every job pushed to Odoo uses a new anonymous
sessions, which creates a tremendous amount of session files.

Using a requests.session means that the session will keep the same
cookie once it receives the first response.

There is 2 problems with the usage of a session though:

* requests.Session is not proved to be threadsafe [0]
Anyway as they are all anonymous, if we lose or mix up odoo sessions at some point it should work as well.
* the main issue is that we don't wait for an answer so until we have a
requests responding before the timeout we won't have any session id

The second issue is resolved in the last commit with a pre-request to obtain a cookie.
When we run a job, we don't wait for the answer. So when we don't have
any cookie yet, we send a GET on a very fast controller route which will
return a cookie, used for all the subsequents calls to runjob.

This effectively prevent to create one session file per job.

I tested it on my side on a large amount of jobs, but extra testing could be helpful. 
What I did not test:
* odoo on a different host / with auth.
* keep it running over a long span of time

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -68,6 +68,18 @@ class RunJobController(http.Controller):
             self.job_storage_class(session).store(job)
         _logger.debug('%s done', job)
 
+    @http.route('/connector/session', type='http', auth="none")
+    def session(self):
+        """ Used by the jobrunner to spawn a session
+
+        The connector jobrunner uses anonymous sessions when it calls
+        ``/connnector/runjob``.  To avoid having thousands of anonymous
+        sessions, before running jobs, it creates a ``requests.Session``
+        and does a GET on ``/connector/session``, providing it a cookie
+        which will be used for subsequent calls to runjob.
+        """
+        return ''
+
     @http.route('/connector/runjob', type='http', auth='none')
     def runjob(self, db, job_uuid, **kw):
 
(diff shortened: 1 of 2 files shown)


sbidoul:
@guewen thanks for tackling this issue!

Code looks good. I'm a bit scared of the shared session object. But from reading around it looks like it should work. So not sure if it would be worth spinning a ThreadPool.

guewen:
About the session shared between threads, my feeling is that with the usage we have here, it should not be an issue.
Also the system is pretty tolerant, from my tests if we use an anonymous session id that does no longer exist on the odoo side, it will be accepted anyway. If we use no session id, odoo will assign a new one which is not a problem if it happens only sparingly.

connector/controllers/main.py
```diff
@@ -68,6 +68,10 @@ def _try_perform_job(self, session_hdl, job):
             self.job_storage_class(session).store(job)
         _logger.debug('%s done', job)
 
+    @http.route('/connector/session', type='http', auth="none")
+    def session(self):
```
simahawk, line 72: docstring pls ;)

simahawk (approved):
LGTM. I'd add a docstring just to mention why that empty method is there ;)

connector/controllers/main.py
guewen: good catch, added

leemannd (approved):

lmignon (approved):

PR #272, opened 2018-01-09
[10.0] component: Fix component lookup by usage

lmignon:
If more than one component is found for a given usage, return the one
linked to the current collection
 
include #271

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -377,6 +377,12 @@ class WorkContext(object):
                 (self.collection._name, usage, model_name)
             )
         elif len(component_classes) > 1:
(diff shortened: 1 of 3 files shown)


guewen (approved):
I think it makes sense, kind of override of the component for a collection. Thanks

lmignon:
@guewen You're right, that's my use case:
* Component A is registered for a specific usage (but not for a collection). This component provides a generic implementation for any models.
* Component B inherits from Component A and provides a specialized implementation applicable to model X and Y in collection  B

simahawk (approved):
@lmignon I had this mind since a while :) thanks!

PR #274, opened 2018-01-12
[8.0] Not only check for operational error's but also retry InternalError's of Psycopg2's layer

lembregtse:
We had the use-case where we also try/catch exceptions of the ORM layers inside our connector jobs, in rare cases of crossing jobs we could trigger an InternalError on the Psycopg2's layer inside our job. 

The error is "IN_FAILED_SQL_TRANSACTION", we consider this to be a retryable error that could be catched on the same level as operational errors.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -2,7 +2,7 @@ import logging
 import traceback
 from cStringIO import StringIO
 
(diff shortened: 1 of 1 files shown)


guewen (approved):
Thanks! Sounds good

PR #280, opened 2018-03-01
[FIX] component: Add missing argument in logging call

naglis:
Hey,

this adds a missing error string formatting argument in the logging call.

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -250,7 +250,7 @@ class WorkContext(object):
                 _logger.error(
                     'No component registry for database %s. '
                     'Probably because the Odoo registry has not been built '
-                    'yet.'
+                    'yet.', dbname
                 )
                 raise
         self._propagate_kwargs = [


simahawk (approved):

guewen:
Of course :no_mouth:
Out of curiosity, did you actually hit this line at runtime or found the error by reading the code?

guewen (approved):

guewen:
The failed build is only due to the docs build. The rest is passing.

naglis:
@guewen yes, I've managed to hit it when running the tests using [`TransactionComponentRegistryCase`](https://github.com/OCA/connector/blob/37dfd976cdbfd50dbb19b175d8b48ed1c6bf5fbe/component/tests/common.py#L194-L195).

guewen:
Thanks for the feedback, I guess it's more likely to happen when writing tests indeed.

PR #284, opened 2018-03-29
Fix cachetools dependency version

guewen:
The current version of cachetools is 2.0.1, so > won't work.
The dependency in the requirements.txt is >=2.0.1

PR #285, opened 2018-03-30
Fix cachetools dependency version and docs

simahawk:
replaces https://github.com/OCA/connector/pull/284 as Guewen is on holidays :)

### DIFF
--- a/connector/doc/api/api_backend.rst
+++ b/connector/doc/api/api_backend.rst
@@ -9,7 +9,7 @@ Backend Model
 *************
 
 
-.. autoclass:: connector.backend_model.ConnectorBackend
+.. autoclass:: connector.models.backend_model.ConnectorBackend
    :members:
    :undoc-members:
    :show-inheritance:
@@ -20,7 +20,7 @@ Backend Model
 Binding Model
 *************
 
-.. autoclass:: connector.backend_model.ExternalBinding
+.. autoclass:: connector.models.backend_model.ExternalBinding
    :members:
    :undoc-members:
    :show-inheritance:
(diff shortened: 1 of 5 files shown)


simahawk:
1st error is gone \o/ still some docs errors

simahawk:
I don't why it fails like this. 
```bash
The command "if [ "$DOCS" != "1" ] ; then travis_run_tests; fi" exited with 0.
6.19s$ if [ "$DOCS" == "1" ] ; then ./.travis_build_doc; fi
Running Sphinx v1.7.2
loading translations [en]... done
making output directory...
loading pickled environment... not yet created
loading intersphinx inventory from http://www.odoo-magento-connector.com/objects.inv...
intersphinx inventory has moved: http://www.odoo-magento-connector.com/objects.inv -> http://odoo-magento-connector.com/objects.inv
loading intersphinx inventory from http://docs.python.org/objects.inv...
intersphinx inventory has moved: http://docs.python.org/objects.inv -> https://docs.python.org/3/objects.inv
loading intersphinx inventory from https://www.odoo.com/documentation/10.0/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 17 source files that are out of date
updating environment: 17 added, 0 changed, 0 removed
reading sources... [  5%] api/api_backend
Recursion error:
maximum recursion depth exceeded
```

guewen:
I fails with a recursion error when trying to autodocument odoo fields, I don't know why, but anyway I don't think we need to list the model fields in the doc. Proposal here: #296 (it replaces this one).

Issue #286, opened 2018-05-15
[8.0, NEW] Sending mails by job


Olivier-LAURENT:
We plan to write a new module to replace current `Email Queue Manager` cron by a mechanism driven by the connector.

The goal is essentially to benefit of a parallelism when sending mails through the smtp server. 

A first approach is: as soon as an outgoing `mail.mail` is created, a `job` responsible of its sending is also created. Mainly this job will call the native `send` method but with only one id. Depending on worker/runner configuration, such jobs could therefore be executed simultaneously.

For instance, together with the `mass_mailing` module that can generate in `mail_mail` resource thousands mails, this module could accelerate the last step, i.e. the physical mails sending, that most often takes a while.

Furthermore if a priority is specified in the calling context, it could be passed to the job, for example, to try to not penalize a small mass mailing while a big ones is in progress.

Of course an attention should assume that the job does not resend an already sent mail (case of auto_send mail) nor fail if the mail is already deleted (case of auto_delete mail).

Later this module would be migrated in the `queue` repository but for now we need such an improvement for an Odoo 8.0 deployment.

Do not hesitate to make any kind of recommendations or suggestions. Thx in advance.

PR #287, opened 2018-05-23
[ADD] New module: mail_connector_queue

andreparames:
This is a proposed module that implements #286.

It creates a new field `priority` in `mail.mail` that is passed to the queue job.

It could be interesting to make the `ChannelJob` extendable and injectable, to allow for custom rankings of jobs (that is, override [this method](https://github.com/OCA/connector/blob/8.0/connector/jobrunner/channels.py#L226)). For example, a custom `MassMailChannelJob` could ignore the date of creation, shuffling old and new jobs without having to specify a specific priority for each batch.
Should I make a proposal for this change?


mail_connector_queue/README.rst
```diff
@@ -0,0 +1,60 @@
+.. image:: https://img.shields.io/badge/licence-AGPL--3-blue.svg
+   :target: http://www.gnu.org/licenses/agpl-3.0-standalone.html
+   :alt: License: AGPL-3
+
+====================
+Mail Connector Queue
+====================
+
+This module uses the Connector Queue to send any emails generated by Odoo,
+by creating a queue job for each one.
```
sbidoul: Can you please put this in readme/DESCRIPTION.rst according to https://github.com/OCA/maintainer-tools/pull/339

mail_connector_queue/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Configuration
+=============
+
+The cron task `Email Queue Manager` should be disabled, as it conflicts with
+this module.
```
sbidoul: s/should/must/

Can you disable it when installing this module?

Please put this in readme/CONFIGURE.rst

mail_connector_queue/README.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Usage
+=====
+
```
sbidoul: Here (in readme/USAGE.rst), explain that jobs are created automatically (and when) and also explain what the priority field does

mail_connector_queue/README.rst
```diff
@@ -0,0 +1,60 @@
...
+Images
+------
+
+* Odoo Community Association: `Icon <https://github.com/OCA/maintainer-tools/blob/master/template/module/static/description/icon.svg>`_.
+
+Contributors
+------------
+
+* André Paramés (https://acsone.eu) <github@andreparames.com>
```
sbidoul: Put this line in readme/CONTRIBUTORS.rst

mail_connector_queue/job.py
```diff
@@ -0,0 +1,22 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openerp.addons.connector.event import on_record_create
+from openerp.addons.connector.queue.job import job
+
+
+@job
+def send_mail(session, mail_id):
+    mail = session.env['mail.mail'].browse(mail_id)
+    if mail.state == 'outgoing':
+        mail.send(auto_commit=True, raise_exception=True)
```
sbidoul: since we are sending one mail, auto_commit is probablly not necessary

mail_connector_queue/job.py
```diff
@@ -0,0 +1,22 @@
...
+    else:
+        return "Not in Outgoing state, ignoring"
```
sbidoul: is this returned text displayed somewhere?

mail_connector_queue/README.rst
sbidoul: A word about why this is useful would be nice too. Such as "This enables, for instance, running high volume mass mailing with low priority, to avoid blocking regular notifications.".

sbidoul (commented):
A few minor remarks, otherwise looks good to me. I love how simple it is :)

pedrobaeza:
Should we ask for new README structure in these new modules for not doubling the work?

sbidoul:
About extending the job priority computation I'm not sure. You have access to very little information when sorting jobs (ie not even a registry). Also this comparison is meant to be extremely fast as we are managing in-memory priority queues that could contains hundreds of thousands entries.

sbidoul:
@pedrobaeza yes we should ask for the new readme structure for new modules or PR touching the README. I'll try to setup the nightly run tonight.

mail_connector_queue/README.rst
andreparames: > Can you disable it when installing this module?

I considered it, but the record has `noupdate=1`; of course I can always override it using SQL, but should we?

mail_connector_queue/job.py
andreparames: Yes, it's shown in the Job form.

mail_connector_queue/README.rst
pedrobaeza: A post_init_hook is the best solution.

sbidoul: > should we?

I think so. This module has no know downside compared to the cron, and things would work poorly if both compete for sending mails. So yes, disable the cron in a post init hook (no need for SQL IMO).

andreparames:
> About extending the job priority computation I'm not sure. You have access to very little information when sorting jobs (ie not even a registry). Also this comparison is meant to be extremely fast as we are managing in-memory priority queues that could contains hundrends of thousands entries.

Fair enough, though in this case the suggestion was to make it depend on *less* information, specifically like this:

```diff
--- a/connector/jobrunner/channels.py
+++ b/connector/jobrunner/channels.py
@@ -231,7 +231,6 @@ class ChannelJob(object):
         else:
             return (cmp(self.eta, other.eta) or
                     cmp(self.priority, other.priority) or
-                    cmp(self.date_created, other.date_created) or
                     cmp(self.seq, other.seq))
```

sbidoul:
Why would you want to remove the comparison on job creation date? If there is no (or same) priority they run FIFO, that's what we want, no?

andreparames:
The idea (just a suggestion) was to alleviate the problem reported in the issue #286 (big newsletters preventing smaller ones from being sent) without forcing the user to know if there's a big newsletter already being sent and manually set the priority.

mail_connector_queue/job.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Copyright 2018 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+from openerp.addons.connector.event import on_record_create
+from openerp.addons.connector.queue.job import job
+
+
+@job
+def send_mail(session, mail_id):
+    mail = session.env['mail.mail'].browse(mail_id)
```
sbidoul: don't we need to handle the (admitedly edge) case where mail_id has been deleted, and do nothing in that case?

guewen:
I have the feeling that allowing custom comparison would add too much complexity for no strong added value. Also I'm not sure about "alleviate the problem [...] big newsletters preventing smaller ones from being sent", at some point if you rely on luck with a shuffled list, it might help, or not at all. If priority is too abstract for users, we could maybe let them choose a selection on the newsletter such as "Low priority", "High priority" and behind the hood set a priority, maybe a different channel for both.

sbidoul:
@andreparames there is also a `mark_outgoing` method on mail.mail that you may want to overwrite to create a job.

And perhaps, in the create overwrite, do not create a job if state is in vals with a value other than outgoing.

mail_connector_queue/README.rst
simahawk: and restore it via `uninstall_hook` ;)

mail_connector_queue/__openerp__.py
```diff
@@ -0,0 +1,21 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+{
+    'name': 'Mail Connector Queue',
+    'summary': """
+        Email Connector Queue""",
+    'version': '8.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA',
```
simahawk: https://github.com/OCA/connector

mail_connector_queue/job.py
simahawk: translatable?

andreparames: @simahawk I considered it, but using which language? I'm not sure it makes sense to use the language of the user sending the email, since it's essentially an internal message. Arguably it should be translated to the language of the user seeing the job form, but that strikes me as adding unnecessary complexity.

mail_connector_queue/job.py
andreparames: Well, it won't do anything now :smile: but I agree, it's clearer for the user.

mail_connector_queue/README.rst
andreparames: @simahawk is there a customary place to store the previous state? Or should I just restore to active regardless?

mail_connector_queue/job.py
simahawk: hmm, indeed this is going to be stored and all in all is just for advanced users: ignore me :stuck_out_tongue_winking_eye:

mail_connector_queue/README.rst
simahawk: no way AFAIK. But we assume that we restore basic odoo behavior, no?

sbidoul:
@andreparames don't forget to push the new files.

mail_connector_queue/readme/USAGE.rst
```diff
@@ -0,0 +1,8 @@
+A queue job is created automatically when a new ``mail.mail`` record is
+created in the ``outgoing`` state. A priority may be defined in the
+``mail.mail`` record, which will be used as the job priority.
+
+
+.. image:: https://odoo-community.org/website/image/ir.attachment/5784_f2813bd/datas
+   :alt: Try me on Runbot
+   :target: https://runbot.odoo-community.org/runbot/102/8.0
```
sbidoul: remove this runbot link, it will be added automatically

sbidoul: remove DESCRIPTION **.md**

mail_connector_queue/readme/USAGE.rst
```diff
@@ -0,0 +1,8 @@
+A queue job is created automatically when a new ``mail.mail`` record is
+created in the ``outgoing`` state. A priority may be defined in the
```
sbidoul, line 2: It must be whenever a mail.mail record reaches the outgoing state, not only creation.

See also https://github.com/OCA/connector/pull/287#issuecomment-391604615

sbidoul (changes requested):
A few more

mail_connector_queue/readme/USAGE.rst
andreparames: @sbidoul Should I commit the generated README?

sbidoul: Normally it's not necessary. But since the README generation system is new, it could be nice so we can have an visual check. Make sure to use the latest master branch of OCA/maintainer-tools

mail_connector_queue/readme/USAGE.rst
andreparames: Overriding `mark_outgoing` will probably work for most cases, but isn't it safer to use `on_record_write` instead, in case some module writes directly to the object?

mail_connector_queue/readme/USAGE.rst
pedrobaeza: @sbidoul If I understood well, no README.rst file is needed at all with the new system. Only the fragments. Is that true? We are trying the new system also in https://github.com/OCA/website/pull/464. Can you check for pointing us anything relevant?

mail_connector_queue/job.py
```diff
@@ -18,6 +18,8 @@ def send_mail(session, mail_id):
 
 @on_record_create(model_names='mail.mail')
 def mail_creation(session, model_name, record_id, vals):
+    if vals.get('state') != 'outgoing':
+        return
```
sbidoul: the default is outgoing (I think, to be checked), so `vals.get('state', 'outgoing')`

mail_connector_queue/readme/USAGE.rst
sbidoul: Absolutely catching all writes is safer

andreparames:
I've pushed a commit that also creates a job when the state changes to `outgoing`.

I added a `SELECT .. FOR UPDATE` at the start of the job to avoid race conditions that could lead to the same email being sent twice, in the (unlikely) case that two jobs are created for the same email - one on `create` and another on `write` -, and then executed at the same time by two different workers.

sbidoul:
Is there not such a select for update in the upstream send method the job is calling?

andreparames:
I looked into `mail/mail_mail.py`, but didn't find any locking. My guess is that Odoo relies on the locking of the cron jobs themselves (which do have a `FOR UPDATE` lock)

sbidoul:
Ah I thought you where calling the cron method itself.

sbidoul:
Well nevermind, I thought there was such a locking mechanism in upstream Odoo, but I can't find it now.

mail_connector_queue/__openerp__.py
```diff
@@ -0,0 +1,23 @@
...
+{
+    'name': 'Mail Connector Queue',
+    'summary': """
+        Email Connector Queue""",
+    'version': '8.0.1.0.0',
+    'license': 'AGPL-3',
+    'author': 'ACSONE SA/NV,Odoo Community Association (OCA)',
+    'website': 'https://github.com/OCA/connector',
+    'depends': [
+        'connector',
```
Olivier-LAURENT, line 14: Hi @andreparames 
The module needs also a dependency on `mail`.

mail_connector_queue/hooks.py
```diff
@@ -0,0 +1,19 @@
...
+    if cron:
+        cron.write({'active': active})
+
+
+def post_init_hook(cr, registry):
+    _set_mail_cron_state(cr, registry, active=False)
+
+
+def uninstall_hook(cr, registry):
+    _set_mail_cron_state(cr, registry, active=False)
```
Olivier-LAURENT: Did you mean here `active=True` ?

Olivier-LAURENT (commented):
Hi @andreparames 
Thx for this module.
IMO two small changes are needed.

andreparames:
Thanks @sbidoul @simahawk @Olivier-LAURENT for the review.

sbidoul (approved):

sbidoul:
Thanks, great job! Looking forward to see how it behaves in production.

simahawk (approved):

guewen (approved):

guewen:
Thanks!

PR #288, opened 2018-05-23
[IMP] Jobrunner as a worker

StefanRijnhart:
We have some memory issues with the job runner, especially when there are 30000 pending jobs (and 30000 ChannelJobs in the uuid registry). The biggest problem is that the job runner lives as a thread in the main process, so that when a new worker is forked, its memory is forked into it. If your soft process memory limit is not outrageous, it could be that it is exceeded at the start of the worker lifecycle so that you get into a loop in which each new worker is retired after one request. This PR suggests as a solution to let the jobrunner live in its own worker.

### DIFF
--- a/connector/jobrunner/__init__.py
+++ b/connector/jobrunner/__init__.py
@@ -61,32 +61,57 @@ class ConnectorRunnerThread(Thread):
         self.runner.stop()
 
 
+class WorkerJobRunner(server.Worker):
+    """ Jobrunner workers """
+
+    def __init__(self, multi):
+        super(WorkerJobRunner, self).__init__(multi)
+        self.watchdog_timeout = None
+        port = os.environ.get('ODOO_CONNECTOR_PORT') or config['xmlrpc_port']
+        base_url = port and 'http://localhost:%s' % port \
+                   or 'http://localhost:8069'
+        channels = _channels()
+        self.runner = ConnectorRunner(base_url, channels or 'root:1')
+
+    def sleep(self):
+        pass
+
+    def signal_handler(self, sig, frame):
+        _logger.debug("WorkerJobRunner (%s) received signal %s", self.pid, sig)
+        super(WorkerJobRunner, self).signal_handler(sig, frame)
+        self.runner.stop()
+
+    def process_work(self):
+        _logger.debug("WorkerJobRunner (%s) starting up", self.pid)
+        time.sleep(START_DELAY)
+        self.runner.run()
+
+
 runner_thread = None
 
-orig_prefork_start = server.PreforkServer.start
-orig_prefork_stop = server.PreforkServer.stop
+orig_prefork__init__ = server.PreforkServer.__init__
+orig_prefork_process_spawn = server.PreforkServer.process_spawn
+orig_prefork_worker_pop = server.PreforkServer.worker_pop
 orig_threaded_start = server.ThreadedServer.start
 orig_threaded_stop = server.ThreadedServer.stop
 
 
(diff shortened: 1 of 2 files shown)


sbidoul:
@StefanRijnhart interesting. I can't say I decided against it :) Sounds like a good idea.

StefanRijnhart:
Good to know! We'll be testing this and I'll report back.

sbidoul:
Might be I wanted to make it work in the threaded server too then did not push it further.

guewen:
> The biggest problem is that the job runner lives as a thread in the main process, so that when a new worker is forked, its memory is forked into it. If your soft process memory limit is not outrageous, it could be that it is exceeded at the start of the worker lifecycle so that you get into a loop in which each new worker is retired after one request.

Make sense, very interesting analysis, I think we have hit this issue a few times. Looking forward to the results

hbrunn:
good idea. But I think we need to keep the thread for threaded servers, and do some self inspection on what version to start.

StefanRijnhart:
@hbrunn did I break that? I left the bits concerning openerp.service.server.ThreadedServer intact I believe.

hbrunn:
thanks for pointing that out, I expected to see some kind of switch statement somewhere, but that happens in core of course and here we just monkey patch the functions being called as result of the conditional. Sorry for the noise

StefanRijnhart:
This has been running stable on one of our development environments running more than 13000 jobs without a restart in the last couple of days.

sbidoul:
@StefanRijnhart have you tested it waits for running jobs to finish on regular Odoo shutdown?

StefanRijnhart:
I'll double check but those jobs are run by other workers so I don't expect a problem there.

StefanRijnhart:
You are on to something. Odoo would not shutdown properly when jobs were being run. I'm now handling SIGINT but I need to do some more testing.

nilshamerlinck:
FYI this improvement would also help in another edge case:
- when notified of a burst of jobs to process, the jobrunner will create a new thread for each job, until reaching maximum defined capacity
- even if these threads don't wait for the request to complete, their simultaneous creation temporarily requires the allocation of more memory to the main process
- that memory increase is never released back to the system
- new workers will start their life with a copy of that extended memory, that might already exceed their soft limit as described above

So definitely an interesting PR! :+1:

sbidoul:
> that memory increase is never released back to the system

How so?

nilshamerlinck:
I was surprised as well when I faced that issue. Sounds like a memory leak but this is actually how memory management works under Linux:

> [processes have a] high-water-mark behavior, where peak memory usage determines the memory usage for the remainder of execution, regardless of whether that memory is actively being used

Source: https://blog.fugue.co/2017-03-06-diagnosing-and-fixing-memory-leaks-in-python.html

StefanRijnhart:
I believe that to be the cause of our memory issues as well (rather than the ChannelJobs). When running 37 jobs concurrently, the memory of our WorkerJobRunner is always around 3.5GB, which is fine if this memory is not forked into regular workers.

StefanRijnhart:
This has been running reliably on all of our intensively used user acceptance systems since the last modification on 29th of May, with no problems on stopping or restarting.

guewen (approved):
I love this :) Thanks!
Can we move forward & merge?

sbidoul:
So now that the JobRunner is in it's own worker it could be gevent'ed to we don't need to create threads to http request jobs. Just a wild idea :)

sbidoul (approved):

guewen:
travis failure is in transifex, unrelated

PR #290, opened 2018-05-28
[8.0] mail_connector_queue: rename field & fetch priority from context

andreparames:
These commits make the field purpose more explicit, and allow passing the priority using the context, as suggested by the original issue #286 but not implemented in the previous PR.

### DIFF
--- a/mail_connector_queue/models/mail_mail.py
+++ b/mail_connector_queue/models/mail_mail.py
@@ -9,4 +9,4 @@ class MailMail(models.Model):
 
     _inherit = 'mail.mail'
 
-    priority = fields.Integer()
+    mail_job_priority = fields.Integer(oldname='priority')
--- a/mail_connector_queue/README.rst
+++ b/mail_connector_queue/README.rst
@@ -37,6 +37,15 @@ A queue job is created automatically when a new ``mail.mail`` record is
 created in the ``outgoing`` state. A priority may be defined in the
 ``mail.mail`` record, which will be used as the job priority.
 
+Changelog
+=========
+
+8.0.1.0.1 (2018-05-28)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* Priority field renamed to ``mail_job_priority``
+* Priority can now be passed using the context
+
 Bug Tracker
 ===========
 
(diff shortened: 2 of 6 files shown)


simahawk (changes requested):
version upgrade will land once merged.
Still, would you mind adding the changelog file in `readme`?

I'd say we should have an entry w/ "Unreleased" name and then when the version is bumped we can set the final number. @sbidoul @guewen ?

sbidoul:
:+1: for the changelog and versioning approach

guewen (approved):

andreparames:
Thanks, I've updated the changelog.

guewen (changes requested):
The test must be updated, it's failing

andreparames:
Tests fixed, sorry for the delay.

simahawk (approved):

mail_connector_queue/models/mail_mail.py
```diff
@@ -2,11 +2,18 @@
...
 
-    priority = fields.Integer()
+    mail_job_priority = fields.Integer(oldname='priority')
+
+    @api.model
+    def create(self, vals):
+        context_value = self.env.context.get('default_mail_job_priority')
+        if context_value and not vals.get('mail_job_priority'):
+            vals['mail_job_priority'] = context_value
+        return super(MailMail, self).create(vals)
```
sbidoul: @andreparames Is this override actually necessary? I'd say Odoo does exactly that out of the box.

sbidoul: Hm, you need to put in vals so the event handlers see it, right? But then the event handler could do a browse to see what priority has actually been saved. Would that not be simpler? Not sure, I'm not familiar with these event handlers.

andreparames: You're right, I see no reason not to fetch from the record by doing a browse. I'll rewrite.

andreparames:
I've updated the PR to avoid the override of the `mail.mail` class, by reading the priority from the record field instead of the creation values, which still allows the priority to be set using the context (as `default_mail_job_priority`).

sbidoul (approved):
Thanks!

Olivier-LAURENT:
👍

guewen (approved):

PR #292, opened 2018-06-04
Exclude components created by tests from the list of addon's components

guewen:
If components are declared in tests, exclude them from the "components
of the addon" list. If not, when we use the "load_components" method,
all the test components would be loaded.  This should never be an issue
when running the app normally, as the Python tests should never be
executed. But this is an issue when a test creates a test components for
the purpose of the test, then a second tests uses the "load_components"
to load all the addons of the module: it will load the component of the
previous test.

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -454,6 +454,17 @@ class MetaComponent(type):
             super(MetaComponent, self).__init__(name, bases, attrs)
             return
 
+        # If components are declared in tests, exclude them from the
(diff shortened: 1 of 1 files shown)


lmignon (approved):
That makes sense and the useful comment will remember us why these 2 lines are there.

TDu (approved):
LGTM

guewen:
The failure is still the docs (couldn't find the issue yet)...

PR #296, opened 2018-07-16
11.0 Fix docs build and improve documentation of APIs

guewen:
Fix the build of `master` which was due to a long-standing issue in the docs generation (recursion error).

* Sphinx crashes with a recursion error when trying to document Models
fields. Remove the fields from the documentation altogether, they didn't
bring much value anyway
* Went through all the API pages and removed many useless autodoc with a
mix of __all__, :exclude-members: or by defining more precisely what we
want with :autoclass:, :autoatttribute:, ... or simply removing them
* Changed the order of a few autodocs by putting the most useful
classes/functions first
* Improved a few docstrings
* Fix connector docs dotted path to models

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -628,15 +628,19 @@ class AbstractComponent(object, metaclass=MetaComponent):
     _abstract = True
 
     # used for inheritance
-    _name = None
+    _name = None  #: Name of the component
+
+    #: Name or list of names of the component(s) to inherit from
     _inherit = None
 
-    # name of the collection to subscribe in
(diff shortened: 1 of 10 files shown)


guewen:
Here's a copy of the generated docs

[doc.tar.gz](https://github.com/OCA/connector/files/2197088/doc.tar.gz)

simahawk (approved):
LG

guewen:
Approved in 10.0 and fixing the build. I merge

PR #298, opened 2018-07-18
Add RecordLocker component

guewen:
This component locks the records for the duration of the transaction
or raise a RetryableJobError if it can't acquire the lock.

### DIFF
--- a/connector/components/locker.py
+++ b/connector/components/locker.py
@@ -1,0 +1,69 @@
+# Copyright 2018 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+import logging
+import psycopg2
+
+from odoo.addons.component.core import Component
+from odoo.addons.connector.exception import RetryableJobError
+
+_logger = logging.getLogger(__name__)
+
+
+class RecordLocker(Component):
+    """Component allowing to lock record(s) for the current transaction
+
+    Example of usage::
+
+        self.component('record.locker').lock(self.records)
+
+    See the definition of :meth:`~lock` for details.
+    """
+
+    _name = "base.record.locker"
+    _inherit = ["base.connector"]
+    _usage = "record.locker"
+
+    def lock(self, records, seconds=None, ignore_retry=True):
+        """Lock the records.
+
+        Lock the record so we are sure that only one job is running for this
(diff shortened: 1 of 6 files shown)


StefanRijnhart (approved):
Cool!

connector/components/locker.py
```diff
@@ -0,0 +1,69 @@
...
+            _logger.info(
+                "A concurrent job is already working on the same "
+                "record (%s with one id in %s). Job delayed later.",
+                self.model._name,
+                tuple(records.ids),
+            )
+            raise RetryableJobError(
+                "A concurrent job is already working on the same record "
+                "(%s with one id in %s). The job will be retried later."
+                % (self.model._name, tuple(records.ids),),
```
simahawk, line 67: does this string interpolation work? Shouldn't you pass `str(records.ids)`?

simahawk (changes requested):
LG, one question

connector/components/locker.py
StefanRijnhart: No, this works. The string substitution implies the cast to a string type.

simahawk: nope, this works, by chance only when you have 1 var x 1 slot, meaning:
```bash
>>> 'Interpolate this please %s' % (tuple([1,2,3]))
  File "<stdin>", line 1, in <module>
TypeError: not all arguments converted during string formatting
>>> 'Interpolate this please %s' % (tuple([1,]))
'Interpolate this please 1'
```

guewen: @simahawk you are not passing a tuple for the interpolation, the code in the PR does:

```python
>>> 'Interpolate this please %s' % (tuple([1,2,3]),)
'Interpolate this please (1, 2, 3)'
>>> 'Interpolate this please %s' % (tuple([1,]),)
'Interpolate this please (1,)'
```

StefanRijnhart: Yes, but the author shows he is aware of this. Note the extra comment after `tuple(record.ids)`

simahawk: hahahaha I definitely need some rest... tricked by sh*tty comma :sweat_smile: 
sorry for the noise guys :stuck_out_tongue:

simahawk (approved):

lmignon (approved):
Great! (Code review only)

PR #299, opened 2018-07-19
[FIX] Prevent two jobrunners from being active on the same database

StefanRijnhart:
a the same time. Stop the Odoo server if this is detected, as it implies that Odoo itself
is already running.

Running multiple jobrunners causes concurrent transaction errors in the
postgresql database. If the job concerns communication with an outside system,
the same jobs are being run indefinitely, causing integrity problems between
Odoo and the external system if the external method is not reentrant.

### DIFF
--- a/connector/jobrunner/runner.py
+++ b/connector/jobrunner/runner.py
@@ -117,6 +117,7 @@ import logging
 import os
 import re
 import select
+import signal
 import threading
 import time
 
@@ -199,6 +200,15 @@ class Database(object):
         self.conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
         self.has_connector = self._has_connector()
         if self.has_connector:
+            if not self._lock():
+                _logger.exception(
+                    "Jobrunner lock was not granted. Is there another "
+                    "jobrunner active in database %s? Trying to stop Odoo.",
+                    self.db_name)
+                os.kill(openerp.service.server.server.pid, signal.SIGINT)
+                raise RuntimeError(
+                    'Jobrunner lock was not granted in database %s' % db_name)
+
             self.has_channel = self._has_queue_job_column('channel')
             self._initialize()
 
@@ -222,6 +232,14 @@ class Database(object):
                 else:
                     raise
             return cr.fetchone()
+
+    def _lock(self):
+        """ Request a session lock on the database. The session lock will be
+        released when self.conn is deleted. If the lock is not granted, there
+        is already a jobrunner running in this database. """
+        with closing(self.conn.cursor()) as cr:
+            cr.execute("select pg_try_advisory_lock(hashtext('jobrunner'))")
+            return cr.fetchone()[0]
 
     def _has_queue_job_column(self, column):
         if not self.has_connector:


StefanRijnhart:
This is how this code gracefully errors out
```
2018-07-19 07:32:42,810 UTC 15253 INFO odoo openerp.addons.connector.jobrunner.runner: initializing database connections
2018-07-19 07:32:42,824 UTC 15253 ERROR odoo openerp.addons.connector.jobrunner.runner: Jobrunner lock was not granted. Is there another jobrunner active in database odoo? Trying to stop Odoo.
2018-07-19 07:32:42,826 UTC 15232 INFO odoo openerp.service.server: Stopping gracefully
2018-07-19 07:32:42,827 UTC 15253 DEBUG odoo openerp.addons.connector.jobrunner: WorkerJobRunner (15253) received signal 2
2018-07-19 07:32:42,827 UTC 15253 INFO odoo openerp.addons.connector.jobrunner.runner: graceful stop requested
2018-07-19 07:32:42,825 UTC 15253 ERROR odoo openerp.addons.connector.jobrunner.runner: exception: sleeping 5s and retrying
  File "/home/odoo/odoo/parts/connector/connector/jobrunner/runner.py", line 385, in run
    self.initialize_databases()
  File "/home/odoo/odoo/parts/connector/connector/jobrunner/runner.py", line 327, in initialize_databases
    db = Database(db_name)
  File "/home/odoo/odoo/parts/connector/connector/jobrunner/runner.py", line 205, in __init__
    'Jobrunner lock was not granted in database %s' % db_name)
RuntimeError: Jobrunner lock was not granted in database odoo
2018-07-19 07:32:42,827 UTC 15251 INFO odoo openerp.service.server: Worker (15251) exiting. request_count: 0, registry count: 1.
2018-07-19 07:32:42,828 UTC 15249 INFO odoo openerp.service.server: Worker (15249) exiting. request_count: 0, registry count: 1.
2018-07-19 07:32:42,829 UTC 15250 INFO odoo openerp.service.server: Worker (15250) exiting. request_count: 0, registry count: 1.
2018-07-19 07:32:47,832 UTC 15253 INFO odoo openerp.addons.connector.jobrunner.runner: stopped
2018-07-19 07:32:47,833 UTC 15253 INFO odoo openerp.service.server: Worker (15253) exiting. request_count: 0, registry count: 1.
```

FYI we have multiple Odoo instances across several systems, one of which is configured for starting the jobrunner. If the deployment system glitches out, the jobrunner is started on multiple instances and everything goes to pot.

connector/jobrunner/runner.py
```diff
@@ -194,6 +195,15 @@ def __init__(self, db_name):
         self.conn.set_isolation_level(ISOLATION_LEVEL_AUTOCOMMIT)
         self.has_connector = self._has_connector()
         if self.has_connector:
+            if not self._lock():
+                _logger.exception(
+                    "Jobrunner lock was not granted. Is there another "
+                    "jobrunner active in database %s? Trying to stop Odoo.",
+                    self.db_name)
+                os.kill(openerp.service.server.server.pid, signal.SIGINT)
```
PieterPaulussen, line 208: Is it possible that this is not executed correctly? Should we not do a try catch around this and perhaps take additional actions to kill Odoo entirely?

PieterPaulussen (changes requested):

connector/jobrunner/runner.py
StefanRijnhart: That's what the raise is for after this. If we cannot kill Odoo for some reason, the exception is raised, the jobrunner sleeps for a little while and tries again. If the main Odoo process does not respond to SIGINT repeatedly, then you probably have bigger problems.

StefanRijnhart: BTW the call itself, even if it does not guarantee that Odoo itself is stopped will not easily raise an exception so try/catch would be void.

PieterPaulussen: Ok, great! Thanks for the clarification.

PieterPaulussen (approved):

sbidoul:
LGTM. Travis error seems unrelated.

guewen:
I like the idea. I have a concern about it though, more than once we experimented some odoo transactions hanging unclosed after shutdown of odoo (perhaps forced kills?), which means the advisory lock would stay around and prevent the restart of any jobrunner until the transaction is killed (timeout). Did you experienced anything like that?

StefanRijnhart:
Sorry for the late reply, but having run this for two years on a high traffic installation using locks here and for other cases, I can safely say that I have never seen an advisory lock linger after Odoo shuts down ;-)

guewen:
Cool!

/ocabot merge minor

PR #300, opened 2018-07-20
[FIX] Get a session with a database

StefanRijnhart:
If a job is executed using a session without a database, Odoo will always connect to the postgreql database for a database list. This is actually expensive because Odoo does not reuse the psycopg2 connection to the postgresql database so every job that is being run leads to a new database connection that is only used once for nothing.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -68,18 +68,6 @@ class RunJobController(http.Controller):
             self.job_storage_class(session).store(job)
         _logger.debug('%s done', job)
 
-    @http.route('/connector/session', type='http', auth="none")
-    def session(self):
-        """ Used by the jobrunner to spawn a session
-
-        The connector jobrunner uses anonymous sessions when it calls
-        ``/connnector/runjob``.  To avoid having thousands of anonymous
-        sessions, before running jobs, it creates a ``requests.Session``
-        and does a GET on ``/connector/session``, providing it a cookie
-        which will be used for subsequent calls to runjob.
-        """
-        return ''
-
     @http.route('/connector/runjob', type='http', auth='none')
     def runjob(self, db, job_uuid, **kw):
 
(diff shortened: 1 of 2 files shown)


StefanRijnhart:
This is actually untested because I don't use the connector on 9.0 but I wanted to have a discussion about this before I backport the whole session feature to 8.0.

sbidoul:
Nothing shocking at first glance. LGTM.

StefanRijnhart:
I think @guewen opted for a 'lightweight' session method on purpose so I'd love to hear from him if this still counts as lightweight or if there are other problems with it.

guewen:
> I think @guewen opted for a 'lightweight' session method on purpose so I'd love to hear from him if this still counts as lightweight or if there are other problems with it.

Overlooked the db handling, that is the reason. 
Looks good to me!

guewen (approved):

simahawk (approved):
build errors related to linting and docs

rmoehn:
Since I need this feature for Odoo 8, I will backport it. Let me know if someone else is already working on that, so we don't do duplicate work.

lmignon:
@StefanRijnhart Your code LGTM. Did you have the opportunity to test it?

StefanRijnhart:
We are actually running this in 8.0, not 9.0 so I can post the backport when I'm behind a screen in a couple of days

lmignon:
Thank you for the follow-up @StefanRijnhart

rmoehn:
@StefanRijnhart, if you won't get around to doing this, please tell us, so I know that I have to stop waiting and start working on it myself.

StefanRijnhart:
@rmoehn here you go: https://github.com/OCA/connector/pull/303. I could not take a clear cherry-pick from our custom branch so it would be good if you could test that version.

rmoehn:
Thanks a lot! I will give it a go on Monday.

guewen:
I could run a successful test with it. Thanks :)

PR #301, opened 2018-07-23
call super().unlink() after the binding has been deleted

damdam-s:
beenn deleted

### DIFF
--- a/connector/producer.py
+++ b/connector/producer.py
@@ -44,9 +44,9 @@ class Base(models.AbstractModel):
 
     @api.multi
     def unlink(self):
-        result = super(Base, self).unlink()
         if is_module_installed(self.env, 'connector'):
             if on_record_unlink.has_consumer_for(self.env, self._name):
                 for record_id in self.ids:
                     on_record_unlink.fire(self.env, self._name, record_id)
-        return result
+
+        return super(Base, self).unlink()


Tonow-c2c (approved):
LGTM

leemannd (approved):

gurneyalex (commented):
I'm pretty sure you still have a race condition here : the execution of the job is asynchronous I think, and can happen after the call to unlink? In that case you still have an error.

guewen:
> I'm pretty sure you still have a race condition here : the execution of the job is asynchronous I think, and can happen after the call to unlink? In that case you still have an error.

At this point there is no job involved, this is the part handling the events propagation, which is all synchronous. The issue here is that when listeners of the event are triggered, the record has already been dropped, so they can't do anything with it.

It was correct in 9.0:

https://github.com/OCA/connector/blob/caee9b08818f550d18bf0193e21ae3abb3d3c632/connector/producer.py#L81-L89

And in 11.0 with the new events 

https://github.com/OCA/connector/blob/71b9d44f06658631367a8843144fdde6b77915ad/component_event/models/base.py#L106-L111

Regarding the jobs, as they are committed in the same transaction, they won't start before in any case.

guewen (approved):
Thanks!

lmignon (approved):

PR #303, opened 2018-08-17
Fix/8.0/session with database

StefanRijnhart:
Backport of Guewen's session mechanism from 9.0 with my modification from #300.

If a job is executed using a session without a database, Odoo will always connect to the postgreql database for a database list. This is actually expensive because Odoo does not reuse the psycopg2 connection to the postgresql database so every job that is being run leads to a new database connection that is only used once for nothing.

### DIFF
--- a/connector/controllers/main.py
+++ b/connector/controllers/main.py
@@ -68,18 +68,6 @@ class RunJobController(http.Controller):
             self.job_storage_class(session).store(job)
         _logger.debug('%s done', job)
 
-    @http.route('/connector/session', type='http', auth="none")
-    def session(self):
-        """ Used by the jobrunner to spawn a session
-
-        The connector jobrunner uses anonymous sessions when it calls
-        ``/connnector/runjob``.  To avoid having thousands of anonymous
-        sessions, before running jobs, it creates a ``requests.Session``
-        and does a GET on ``/connector/session``, providing it a cookie
-        which will be used for subsequent calls to runjob.
-        """
-        return ''
-
     @http.route('/connector/runjob', type='http', auth='none')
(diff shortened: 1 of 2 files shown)


rmoehn:
This works for me.

It's unusual, though, to leave no space between a commit message's subject line and body.

StefanRijnhart:
First time I heard that ever, but then I'm a self taught man and may suffer from ideosyncrasies.

rmoehn:
Yeah, I have those too. Like standing on one leg in front of my desk. But here's a quote from the `git commit` man page.

> Though not required, it’s a good idea to begin the commit message with a single short (less than 50 character) line summarizing the change, **followed by a blank line** and then a more thorough description. […]

This is echoed in several texts on Git commit message best practices.

sbidoul:
LGTM 

Travis redness seems to be a tiny rst syntax error in another module.

rmoehn:
A note about something I just stumbled over: If you have a separate server for running jobs, it now needs to `--load=…,web` as well. Before this PR that wasn't needed.

StefanRijnhart:
I adapted the commit message. I can't see how this affects preloading even though I did not know the jobrunner could work without preloading in the first place.

rmoehn:
It always needed `--load=connector`. But after this PR it needs `--load=connector,web`. Otherwise we get a 404 on the login request. I should have expected that, but I didn't and scratched my head for a while. So I wanted to point it out in case someone else is also scratching their head and finds this thread.

Thanks for adapting the commit message.

StefanRijnhart:
Ah, thanks for the clarification! Makes sense now.

guewen (approved):

PR #304, opened 2018-08-21
[FIX] 8.0 travis

StefanRijnhart:


pedrobaeza (approved):
Fast-tracking as the fix is straight forward

PR #305, opened 2018-08-21
[FIX] 9.0 travis

StefanRijnhart:


pedrobaeza (approved):

connector/doc/api/api_backend.rst
```diff
@@ -64,6 +64,5 @@ Backend Models
 ==============
 
 .. automodule:: connector.backend_model
-   :members:
```
lmignon, line 67: @StefanRijnhart Why do you have to remove the `:members:` directive?

StefanRijnhart: The members directive causes the documentation build job to fail with `maximum recursion depth exceeded while calling a Python object`. See https://travis-ci.org/OCA/connector/jobs/406290979

The fix was taken from https://github.com/OCA/connector/commit/3b689760798 which is in the 10.0 branch.

Sorry that I did not copy this info from the commit message to the description of this PR.

lmignon:
Thank you @StefanRijnhart!

PR #310, opened 2018-10-02
Migrate component to version 12.0

guewen:


component/__manifest__.py
```diff
@@ -15,5 +17,6 @@
      'python': ['cachetools'],
  },
  'data': [],
- 'installable': False,
+ 'installable': True,
```
lmignon, line 19: @guewen It should be nice to also have the development status key into the manifest
see https://odoo-community.org/page/development-status

lmignon (approved):
Thank you @guewen A little comment...

component/__manifest__.py
guewen: Right, I was not sure at which moment it should be put, is stable the correct status?

lmignon: IMO yes it's stable!

guewen:
Travis is red only due to the "docs" build which cannot build without the "connector" addon

component/__manifest__.py
```diff
@@ -3,7 +3,9 @@
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
 
 {'name': 'Components',
- 'version': '11.0.1.1.0',
+ 'summary': 'Add capabilities to register and use decoupled components,'
+            ' as an alternative to model classes',
+ 'version': '12.0.1.0.0',
  'author': 'Camptocamp,'
            'Odoo Community Association (OCA)',
  'website': 'https://www.camptocamp.com',
```
simahawk: https://github.com/OCA/connector

component/__manifest__.py
```diff
@@ -15,5 +17,7 @@
      'python': ['cachetools'],
  },
  'data': [],
```
simahawk: drop data

simahawk (approved):
minor glitches

component/__manifest__.py
guewen: done

component/__manifest__.py
guewen: done

lmignon (approved):
Once again, thank you @guewen for this migration. IMO we don't need more to officially release this addon for Odoo 12.

rvalyi (commented):
I approve. Travis is green minus the doc task due to the missing connector module in branch 12.0 at this stage, so okay. Also there is very little dependency on Odoo code so it's normal there is no change needed in the code. Thanks.

rvalyi (approved):

PR #316, opened 2018-11-28
[10.0][FIX] Fix component lookup when no result based on the collection

acsonefho:
**Context**
To create my new `Component` (a json export mapper for my case), I only specify a `_apply_on` and `_usage` (also `_name` and `_inherit` of course).
**Bug**
If I don't specify a `_collection`, my component is ignored.
**Expected behaviour**
If no `_collection` specified on the component, it's because it could be apply for every collections (for models in `_apply_on`)

### DIFF
--- a/component/tests/test_component.py
+++ b/component/tests/test_component.py
@@ -41,10 +41,23 @@ class TestComponent(TransactionComponentRegistryCase):
             _usage = 'for.test'
             _apply_on = ['res.users']
 
+        class ComponentCompanyBase(Component):
+            _name = 'component.company.base'
+            _usage = 'test.mapper'
+
+        class ComponentCompany(Component):
+            # Do not specify a collection
+            _name = 'component.company'
+            _inherit = 'component.company.base'
+            _apply_on = ['res.company']
+            _usage = 'test.mapper'
+
         # build the components and register them in our
         # test component registry
         Component1._build_component(self.comp_registry)
         Component2._build_component(self.comp_registry)
+        ComponentCompanyBase._build_component(self.comp_registry)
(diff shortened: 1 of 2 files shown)


lmignon:
@acsonefho Can you please add a unittest to illustrate the problem? 
Regards,

lmi

acsonefho:
@lmignon I create a unit test to reproduce the issue.

lmignon (approved):
Thank you @acsonefho LGTM
@guewen Can you check this one?

component/tests/test_component.py
```diff
@@ -107,6 +120,19 @@ def test_component_get_by_name_other_model(self):
             self.assertEquals('res.partner', base.work.model_name)
             self.assertEquals('res.users', comp.work.model_name)
 
+    def test_without_collection(self):
+        """
+        Test to reproduce issue when no collection is specified.
+        Bug: when no collection is specified, the component fonction
+        is too restrictive and doesn't find a component.
+        If no component found, the work.component(usage='test.mapper') will
+        raise an exception
```
guewen: The docstring is a bit weird because after this merged, there will be no bug anymore ;) It'd make more sense to only describe the expected behavior, rather than an unexpected one. Something like "For 2 candidate Components without collection, one without _apply_on and another one with, the component with a matching model must be returned"

guewen (approved):
Looks good thanks! The docstring of the test can be improved IMO

component/tests/test_component.py
acsonefho: @guewen Thank you, I change the docstring :)

guewen:
Thanks @OCA-git-bot :laughing:

PR #325, opened 2019-02-25
[ADD] requeue of stucked jobs

mpanarin:
will be moved to the queue repo for version 10-12

### DIFF
--- a/connector/queue/model.py
+++ b/connector/queue/model.py
@@ -22,12 +22,13 @@
 import logging
 from datetime import datetime, timedelta
 
-from openerp import models, fields, api, exceptions, _
+from openerp import _, api, exceptions, fields, models
+from openerp.osv import expression
 
-from .job import STATES, DONE, PENDING, ENQUEUED, FAILED, \
-    OpenERPJobStorage, JOB_REGISTRY
-from ..session import ConnectorSession
 from ..connector import get_openerp_module, is_module_installed
+from ..session import ConnectorSession
+from .job import (DONE, ENQUEUED, FAILED, JOB_REGISTRY, PENDING, STATES,
+                  OpenERPJobStorage)
 
 _logger = logging.getLogger(__name__)
 
@@ -212,6 +213,55 @@ class QueueJob(models.Model):
         )
         jobs.unlink()
         return True
+
+    @api.model
+    def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
+        """Fix jobs that are in a bad states
+        :param in_queue_delta: lookup time in minutes for jobs
+                                that are in enqueued state
+
+        :param started_delta: lookup time in minutes for jobs
+                                that are in enqueued state,
+                                0 means that it is not checked
+        """
+        self._get_stuck_jobs_to_requeue(
+            enqueued_delta=enqueued_delta,
+            started_delta=started_delta
+        ).requeue()
+        return True
+
+    @api.model
+    def _get_stuck_jobs_domain(self, queue_dl, started_dl):
+        domain = []
+        now = fields.datetime.now()
+        if queue_dl:
+            queue_dl = now - timedelta(minutes=queue_dl)
+            domain.append([
+                '&',
+                ('date_enqueued', '<=', fields.Datetime.to_string(queue_dl)),
+                ('state', '=', 'enqueued'),
+            ])
+        if started_dl:
+            started_dl = now - timedelta(minutes=started_dl)
+            domain.append([
+                '&',
+                ('date_started', '<=', fields.Datetime.to_string(started_dl)),
+                ('state', '=', 'started'),
+            ])
+        if not domain:
+            raise exceptions.ValidationError(
+                _("If both parameters are 0, ALL jobs will be requeued!")
+            )
+        return expression.OR(domain)
+
+    @api.model
+    def _get_stuck_jobs_to_requeue(self, enqueued_delta, started_delta):
+        job_model = self.env['queue.job']
+        stuck_jobs = job_model.search(self._get_stuck_jobs_domain(
+            enqueued_delta,
+            started_delta,
+        ))
+        return stuck_jobs
 
 
 class RequeueJob(models.TransientModel):
--- a/connector/queue/queue_data.xml
+++ b/connector/queue/queue_data.xml
@@ -2,6 +2,16 @@
 <openerp>
     <data noupdate="1">
 
+        <record id="ir_cron_connector_queue_garbage_collector" model="ir.cron">
+            <field name="name">Jobs Garbage Collector</field>
+            <field name="interval_number">5</field>
+            <field name="interval_type">minutes</field>
+            <field name="numbercall">-1</field>
+            <field name="model">queue.job</field>
+            <field name="function">requeue_stuck_jobs</field>
+            <field name="args">()</field>
+        </record>
+
         <!-- Connector-related subtypes for messaging / Chatter -->
         <record id="mt_job_failed" model="mail.message.subtype">
             <field name="name">Job failed</field>


sebalix (approved):

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+                            that are in enqueued state
+    """
+    obj = session.env[model_name]
+    obj.treat(
+        obj.get_jobs_to_requeue(
+            in_queue_delta=in_queue_delta,
+            started_delta=started_delta
+        )
+    )
+    return True
```
guewen: Using a job for this will be an issue, because the main reason why we would want to requeue jobs stuck in started or enqueued is because they are stuck and block the queue: this job will likely never be executed. Better by executed directly by the cron

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+            self._name,
+        )
+
+    @api.multi
+    def get_jobs_to_requeue(self, in_queue_delta, started_delta):
+        job_object = self.env['queue.job']
+        now = fields.datetime.now()
+        queue_dl = now - timedelta(minutes=in_queue_delta)
+        started_dl = now - timedelta(minutes=started_delta)
+        sick_jobs = job_object.search(
```
guewen:
```suggestion
        stuck_jobs = job_model.search(
```

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+    def set_cleanup_job(self):
+        session = ConnectorSession.from_env(self.env)
+        cron_checkup.delay(
+            session,
+            self._name,
+        )
+
+    @api.multi
+    def get_jobs_to_requeue(self, in_queue_delta, started_delta):
+        job_object = self.env['queue.job']
```
guewen:
```suggestion
        job_model = self.env['queue.job']
```

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+        now = fields.datetime.now()
+        queue_dl = now - timedelta(minutes=in_queue_delta)
+        started_dl = now - timedelta(minutes=started_delta)
+        sick_jobs = job_object.search(
+            [('date_enqueued', '<=', fields.Datetime.to_string(queue_dl)),
+             ('state', '=', 'enqueued')]
+        )
+        sick_jobs += job_object.search(
```
guewen:
```suggestion
        stuck_jobs += job_model.search(
```

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+            [('date_started', '<=', fields.Datetime.to_string(started_dl)),
+             ('state', '=', 'started')]
+        )
+        return sick_jobs
```
guewen:
```suggestion
        return stuck_jobs
```

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @api.multi
+    def treat(self, jobs):
```
guewen: I don't think we need 4 methods for requeuing stuck jobs. I would keep only `get_jobs_to_requeue` (renamed as `get_stuck_jobs_to_requeue` and `set_cleanup_job` renamed as `requeue_stuck_jobs`.

queue_garbage_collector/models/garbage_collector.py
```diff
@@ -0,0 +1,66 @@
...
+            in_queue_delta=in_queue_delta,
+            started_delta=started_delta
+        )
+    )
+    return True
+
+
+class ConnectorQueueGarbageCollector(models.TransientModel):
+    """Fix jobs that are in a bad state"""
+    _name = "connector.queue.garbage.collector"
```
guewen: No need for a new model, the couple of methods can be added to queue.job

guewen (changes requested):
Thanks for your work so far
We shouldn't have a new addon for this, could you move the methods and the cron in the `queue_job` addon?

queue_garbage_collector/models/garbage_collector.py
mpanarin: maybe we can move it to a separate channel? Or does the queue get blocked completely?

guewen: It depends: if a job is stuck in "started" in the root channel, it blocks other jobs.
It's not worth having this method in a job, keep it simple.

simahawk (changes requested):
quoting Guewen's remarks :)

connector/queue/model.py
```diff
@@ -213,6 +213,39 @@ def autovacuum(self):
...
+        jobs = self.get_stuck_jobs_to_requeue(
+            in_queue_delta=in_queue_delta,
+            started_delta=started_delta
+        )
+        for job in jobs:
+            job.requeue()
+        return True
+
+    @api.multi
+    def get_stuck_jobs_to_requeue(self, in_queue_delta, started_delta):
```
simahawk:
```py
@api.model
def _get_stuck_jobs_to_requeue
```

connector/queue/model.py
```diff
@@ -213,6 +213,39 @@ def autovacuum(self):
...
+        job_model = self.env['queue.job']
+        now = fields.datetime.now()
+        queue_dl = now - timedelta(minutes=in_queue_delta)
+        started_dl = now - timedelta(minutes=started_delta)
+        stuck_jobs = job_model.search(
```
simahawk: what about a single query w/ an OR? :smile:

connector/queue/model.py
```diff
@@ -213,6 +213,39 @@ def autovacuum(self):
...
+                                that are in enqueued state
+
+        :param started_delta: lookup time in minutes for jobs
+                                that are in enqueued state
+        """
+        jobs = self.get_stuck_jobs_to_requeue(
+            in_queue_delta=in_queue_delta,
+            started_delta=started_delta
+        )
+        for job in jobs:
```
simahawk: no need to loop
```py
self._get_stuck_jobs_to_requeue(
    in_queue_delta=in_queue_delta,
    started_delta=started_delta
).requeue()
```

simahawk (changes requested):

mpanarin:
@simahawk @guewen updated

simahawk (approved):

connector/queue/model.py
```diff
@@ -213,6 +213,43 @@ def autovacuum(self):
         jobs.unlink()
         return True
 
+    @api.model
+    def requeue_stuck_jobs(self, in_queue_delta=5, started_delta=180):
```
guewen:
```suggestion
    def requeue_stuck_jobs(self, enqueued_delta=5, started_delta=0):
```
`enqueued_delta` is better than `in_queue_delta` as it reflects the name of the state.
We don't know if users in the wild have "legit" jobs that runs more than 3 hours, so we should probably not set any default value for `started_delta`. A value of 5 minutes for `enqueued_delta` is fine as it should not happen.

Update the docstring to say that 0 means it is not checked

connector/queue/model.py
```diff
@@ -213,6 +213,43 @@ def autovacuum(self):
...
+                '&',
+                ('date_started', '<=', fields.Datetime.to_string(started_dl)),
+                ('state', '=', 'started')]
+
+    @api.model
+    def _get_stuck_jobs_to_requeue(self, in_queue_delta, started_delta):
+        job_model = self.env['queue.job']
+        now = fields.datetime.now()
+        queue_dl = now - timedelta(minutes=in_queue_delta)
+        started_dl = now - timedelta(minutes=started_delta)
```
guewen: Consider that the deltas can be 0, build the domain only with deltas > 0.

guewen (approved):
Thanks :)

PR #327, opened 2019-03-19
[12.0][FIX] component, component_event: tag unittest.TestCase subclasses

naglis:
Since [v12.0](https://github.com/odoo/odoo/commit/b356b190338e3ee032b9e3a7f670f76468965006), direct subclasses of `unittest.TestCase` which do not have :class:`odoo.tests.common.MetaCase` as a meta-class must be explicitly tagged with :method:`odoo.tests.common.tagged`, otherwise they will not be picked up by the test runner. E.g.:

```python
@tagged('standard', 'at_install')
```

### DIFF
--- a/component/tests/common.py
+++ b/component/tests/common.py
@@ -99,7 +99,8 @@ class SavepointComponentCase(common.SavepointCase, ComponentMixin):
         ComponentMixin.setUp(self)
 
 
-class ComponentRegistryCase(unittest.TestCase):
+class ComponentRegistryCase(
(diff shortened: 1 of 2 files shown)


guewen (approved):
Wow, https://github.com/odoo/odoo/commit/b356b190338e3ee032b9e3a7f670f76468965006 breaks my expectations :/ 
Many thanks for the correction :star:

lmignon (approved):
@naglis Thank you  for the fix and the explanation.

simahawk (approved):

PR #340, opened 2019-06-29
[IMP] Add Infor and Voicent connectors

max3903:


guewen (approved):

guewen (changes requested):
Oops, `gh-pages` are regenerated automatically from the code branches.
The file you want to update is https://github.com/OCA/connector/blob/12.0/connector/doc/index.rst

PR #342, opened 2019-07-09
[9.0]] [IMP] Check if a job is already running when enqueing

StephanRozendaal:
When using a setup with multiple servers that are running the connector, a single job can be picked up by multiple servers at the same time, producing unwanted results.

This is solved by adding a where state = 'pending' clause when updating the state of the queue job on enqueuing.


StefanRijnhart:
Do you run workers or jobrunners? Don't you have concurrency errors all the time? I would suggest to run jobrunner on one of the machines and suppress starting the jobrunner or workers on the other machines. Assuming you have a load balancer, you can have jobrunner call the load balancer URL to run the jobs.

PR #356, opened 2019-10-14
[12.0][CHG] connector: Use more permissive licence: AGPL-> LGPL

lmignon:
This change will allow to use these framwork addons to build addons on top of OE  addons.

Even if I would have preferred that this module remain under AGPL license, I can only state that the AGPL license is now beginning to cause problems on some of our projects where the customer's needs bring us to develop connectors that also handle data from company modules.

More generally, I think that the most relevant choice for low-level technical modules in OCA is now the LGPL license.

ping @OCA/connector-maintainers


guewen (approved):
As the modules built on top of `component` and `connector` are generally crossing with all features modules (e.g. you want to expose models including the EE ones or synchronize data including EE models), it makes sense from a technical perspective. It's not with a light heart, but it has to be done, sadly. I prefer these addons to be used in such situations rather than not (or illegally).

simahawk (approved):

jgrandguillaume (approved):
Sounds like a good move to increase the usage of such a low level feature

PR #359, opened 2020-01-02
[FIX] Define model as property on work context

jcoux:
To allow to change the context of model returned by work context,
we need to define model as property.

Like this, we can change the collection environment with new context,
and model will get from contexted environment.


sebalix (approved):
Seems good

guewen (approved):

PR #371, opened 2020-05-25
Recover the v13 new feature of create method with a list of vals on Base

jbaudoux:
Currently, as soon as this module is installed, each loading of multiple records at a time is converted to  a loading of a single record at a time.
I changed the decorator to support vals_list in create


sebalix (approved):

guewen (approved):
Good catch

guewen:
@gurneyalex do you have an idea why the 13.0 runbot builds are all skipped? https://runbot.odoo-community.org/runbot/repo/https-github-com-oca-connector-102

PR #378, opened 2020-09-09
[FIX] connector: restore binder context

danielduqma:
When binder looks for a binding, search is performed indicating `active_test=False` in context, but found binding keeps this key in context affecting later operations.

For example, having an already imported sales order, an update operation is made to add a new line. When excepcions from `sale_exception` are tested, the ones that are disabled are also tested, because as `active_test=False` is in context they are found too.


guewen (approved):
Good catch

danielduqma:
Can it be merged then, please? After this I forward port it to 12.0 and 13.0 branches. Thanks!

Regards

PR #382, opened 2020-10-12
MIGR 14.0 component

bealdav:


guewen:
Thanks for taking care of this!

bealdav:
Thanks for you attention on my PR @guewen . In my first commit here I didn't updates these files but OCA warnings were raised. I understand the purpose and I'm OK to revert but we have to deal with OCA checks

guewen:
I think they are non blocking warnings, they were already there in earlier versions

bealdav:
OK changes done. But the tests'll fail until I add this kind of code in component to (in next commit).
https://github.com/odoo/odoo/blob/14.0/odoo/tests/common.py#L248-L268

It's a hack about python version. I think with have to deal with it in several module in OCA.
I'm very sad than in OCA we don't only support python 3.8 because it's very stable (first version released on 19th November) and now python is on 3.9.

EDIT:
We have this kind of error.
https://travis-ci.org/github/OCA/connector/jobs/735007451#L750-L752

bealdav:
@guewen could you have a look please ?

bealdav:
cc @simahawk @lmignon please could you have a look to the last commit ? I would like to have your opinion. Thanks

component/README.rst
```diff
@@ -0,0 +1,129 @@
...
+
+    This file may contain ONE level of section titles, underlined
+    with the ~ (tilde) character. Other section markers are
+    forbidden and will likely break the structure of the README.rst
+    or other documents where this fragment is included. ]
+
+Next
+~~~~
+
+12.0.1.0.0 (2018-10-02)
```
mourad-ehm, line 77: 14.0.1.0.0 ?

component/README.rst
```diff
@@ -0,0 +1,129 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+* [MIGRATION] from 11.0 branched at rev. 324e006
```
mourad-ehm, line 80: from 11.0 ?

mourad-ehm (approved):

eantones:
@bealdav I created a PR https://github.com/akretion/connector/pull/1 on your branch fixing some migration issues. Please review it.

bealdav:
Thanks for your review. As said @guewen tells me to not update include path. I've updated missing lib and deprecated assertEquals

Thanks to have point me on it

guewen (approved):
Could be worth updating in the readme fragment for the history at which commit of 13.0 it was branched as pointed by @mourad-ehm.

Looks good though, thanks!

bealdav:
Thanks @guewen I'll check this file.
Other topic. What is your opinion concerning maintain different code for different python version. Is that worth it to support older than 3.8 from a pragmatic point of view ?

guewen:
> Other topic. What is your opinion concerning maintain different code for different python version. Is that worth it to support older than 3.8 from a pragmatic point of view ?

IMO we have no choice but to support the same versions as Odoo does.

component/__manifest__.py
```diff
@@ -5,13 +5,18 @@
...
     "author": "Camptocamp," "Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/connector",
     "license": "LGPL-3",
     "category": "Generic Modules",
     "depends": ["base"],
-    "external_dependencies": {"python": ["cachetools"]},
+    "external_dependencies": {
+        "python": [
+            "cachetools",
+            "mock",
```
simahawk: why do we need mock here?

eantones: Because it's used on tests. Do you think that it's not necessary because the tests are not the code itself? Just in the same sense as @guewen mentioned in https://github.com/akretion/connector/pull/2#discussion_r509111937?

bealdav: Ok I see, to be installed and executed in our projects this module doesn't need mock.
Only to be tested it needs it. Probably it'll be tested only on OCA and on developer computer. 
That's what @simahawk means I suppose.

eantones: It's a good point

simahawk: in any case mock is a built-in lib since py3... you won't need it anyway https://docs.python.org/3.5/library/unittest.mock.html

bealdav: Thanks for the info

bealdav:
Ok then can be merged

component/README.rst
bealdav: Fixed in HISTORY.rst

component/README.rst
bealdav: Fixed in HISTORY.rst

bealdav:
Now green you may review https://github.com/OCA/connector/pull/384
Thanks

component/__manifest__.py
guewen: Wasn't this line to remove then?

simahawk: yep, unattended

bealdav: Finally no more mock in this PR @guewen @simahawk

guewen: Thanks!

guewen:
/ocabot merge nobump


Many thanks @bealdav

PR #383, opened 2020-10-12
[FIX][10.0] When searching for bindings, also find inactive ones

thomaspaulb:
If the child record is active (or doesnt have an active field) but the dependent record is inactive, the connector will try to repeatedly create bindings for the parent record. Prevent this by searching bindings with `active_test=False`.

NOTES:

- I considered using [binder.to_external](https://github.com/OCA/connector/blob/10.0/connector/components/binder.py#L40) which already has these features, but then eg. the functionality of specifying the binding_field explicitly is lost.

- I took into account [this fix](https://github.com/OCA/connector/pull/378/files) and restore the original context of the binding after doing `active_test=False`, so that it doesnt propagate for people using the binding.

- Please review. I can forward port this after merge.

### DIFF
--- a/connector/components/synchronizer.py
+++ b/connector/components/synchronizer.py
@@ -317,9 +317,14 @@ class GenericExporter(AbstractComponent):
         wrap = relation._name != binding_model
 
         if wrap and hasattr(relation, binding_field):
-            domain = [('odoo_id', '=', relation.id),
-                      ('backend_id', '=', self.backend_record.id)]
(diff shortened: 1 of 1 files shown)


thomaspaulb:
@NL66278 Do you have any experience with the connector module?

NL66278:
@thomaspaulb I have a few times looked at it, but experienced it as too complicated, layering abstraction on abstraction, making it fairly incomprehensible.

guewen (approved):
Thanks!

PR #384, opened 2020-10-12
MIGR 14.0 component_event

bealdav:
based on https://github.com/OCA/connector/pull/382

There is the same error with missing method in the test


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo14-addon-component_event @ git+https://github.com/OCA/connector@refs/pull/382/head#subdirectory=setup/connector_event
```
eantones: @bealdav Shouldn't be pointing at **component** instead of **component_event** (itself)? Also I think that the subdirectory **setup/connector_event** should be **setup/component**. In summary, I think it should be:

`odoo14-addon-component @ git+https://github.com/OCA/connector@refs/pull/382/head#subdirectory=setup/component`

eantones:
@bealdav  I created a PR https://github.com/akretion/connector/pull/2 on your branch fixing some migration issues. Please review it. Thanks

eantones (changes requested):

eantones (approved):

test-requirements.txt
guewen: component is now merged

bealdav: I removed content here @guewen

guewen (approved):

PR #392, opened 2020-12-16
[13.0] component: small improvements

simahawk:
1. propagate lookup values to _component_match
2. `component()` now accepts `first` kw arg to not fail when several components are found.

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -346,14 +346,31 @@ class WorkContext(object):
             work_context = self.work_on(model_name)
         return component_class(work_context)
 
-    def _lookup_components(self, usage=None, model_name=None):
+    def _lookup_components(self, usage=None, model_name=None, **kw):
         component_classes = self.components_registry.lookup(
             self.collection._name, usage=usage, model_name=model_name
         )
+        matching_components = []
+        for cls in component_classes:
+            try:
+                matching = cls._component_match(
+                    self, usage=usage, model_name=model_name, **kw
+                )
+            except TypeError as err:
+                # Backward compat
+                _logger.info(str(err))
+                _logger.info(
+                    "The signature of %s._component_match has changed. "
+                    "Please, adapt your code as "
+                    "(self, usage=usage, model_name=model_name, **kw)",
+                    cls.__name__,
+                )
+                matching = cls._component_match(self)
+            if matching:
+                matching_components.append(cls)
+        return matching_components
 
-        return [cls for cls in component_classes if cls._component_match(self)]
-
-    def component(self, usage=None, model_name=None):
+    def component(self, usage=None, model_name=None, **kw):
         """ Find a component by usage and model for the current collection
 
         It searches a component using the rules of
@@ -379,7 +396,9 @@ class WorkContext(object):
         if isinstance(model_name, models.BaseModel):
             model_name = model_name._name
         model_name = model_name or self.model_name
-        component_classes = self._lookup_components(usage=usage, model_name=model_name)
+        component_classes = self._lookup_components(
+            usage=usage, model_name=model_name, **kw
+        )
         if not component_classes:
             raise NoComponentError(
                 "No component found for collection '%s', "
@@ -416,7 +435,7 @@ class WorkContext(object):
             work_context = self.work_on(model_name)
         return component_classes[0](work_context)
 
-    def many_components(self, usage=None, model_name=None):
+    def many_components(self, usage=None, model_name=None, **kw):
         """ Find many components by usage and model for the current collection
 
         It searches a component using the rules of
@@ -430,7 +449,9 @@ class WorkContext(object):
         if isinstance(model_name, models.BaseModel):
             model_name = model_name._name
         model_name = model_name or self.model_name
-        component_classes = self._lookup_components(usage=usage, model_name=model_name)
+        component_classes = self._lookup_components(
+            usage=usage, model_name=model_name, **kw
+        )
         if model_name == self.model_name:
             work_context = self
         else:
@@ -664,7 +685,7 @@ class AbstractComponent(object, metaclass=MetaComponent):
         self.work = work_context
 
     @classmethod
-    def _component_match(cls, work):
+    def _component_match(cls, work, usage=None, model_name=None, **kw):
         """ Evaluated on candidate components
 
         When a component lookup is done and candidate(s) have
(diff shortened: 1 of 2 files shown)


component/core.py
```diff
@@ -346,14 +346,18 @@ def component_by_name(self, name, model_name=None):
...
         component_classes = self.components_registry.lookup(
             self.collection._name, usage=usage, model_name=model_name
         )
 
-        return [cls for cls in component_classes if cls._component_match(self)]
+        return [
+            cls
+            for cls in component_classes
+            if cls._component_match(self, usage=usage, model_name=model_name, **kw)
+        ]
```
guewen, line 381: Should we degrade nicely in case an implementation have `_component_match` with the old signature?
That would force to have a less simple expression but would be more helpful for users:

```python
matching_components = []
for cls in component_classes:
    try:
        matching = cls._component_match(self, usage=usage, model_name=model_name, **kw)
    except TypeError as err:
        _logger.warn("The signature of %s._component_match must be changed to (self, usage=usage, model_name=model_name, **kw)", cls.__name__)
        matching =  cls._component_match(self)
    if matching:
        matching_components.append(cls)
return matching_components
```

component/core.py
```diff
@@ -400,23 +406,26 @@ def component(self, usage=None, model_name=None):
...
-            raise SeveralComponentError(
-                "Several components found for collection '%s', "
-                "usage '%s', model_name '%s'. Found: %r"
-                % (
-                    self.collection._name,
-                    usage or "",
-                    model_name or "",
-                    component_classes,
+            if first:
+                component_classes = component_classes[:1]
```
guewen: Still a bit puzzled by this after our discussion. This is not deterministic so is it really good?
Also, on the fact that you don't want to use `many_components()[0]`, did you actually measure the cost of the instantiation of the components? 
We shouldn't sacrifice API's understandability if the optimization is negligible.
In this case, I think it makes the interface less clear (different exceptions depending on parameter, not deterministic result).

An alternative would be to add a private method

```python
def _matching_components(self, usage=None, model_name=None, **kw):
        if isinstance(model_name, models.BaseModel):
            model_name = model_name._name
        model_name = model_name or self.model_name
        component_classes = self._lookup_components(
            usage=usage, model_name=model_name, **kw
        )
        if model_name == self.model_name:
            work_context = self
        else:
            work_context = self.work_on(model_name)
        return component_classes, work_context
```

This method can be used in `components` and `many_components`, but also from implementations which can filter again the component_classes and instantiate by themselves (which breaks the current contract that only instantiated components are returned). But again, I have the feeling the instantiation is negligible...

component/core.py
simahawk: yeah, I though about it but them I thought: how many devs had ever customized this method? :sweat_smile: 

I'll apply this change anyway. Safer, less possible headaches.

component/core.py
simahawk: yeah, I was typing something similar (`_query_components` in my case) but then I decided to keep it simple and not change too much what is there today.
Regarding the cost of init: well, you don't know. It really depends on what ppl do in their init if they customized it.
For example: I added a validation of the work ctx attributes whereas if a required attribute is not there an exception is raised. In this case, I don't won't that piece of code to be awaken if is not needed.

component/core.py
guewen: > how many devs had ever customized this method? sweat_smile

I did :laughing:

component/core.py
guewen: > not change too much what is there today.

In my head, adding such parameter is a heavier change than extracting the method `_query_components` :shrug:.

What about the fact that the component returned is not deterministic? What's the idea?

FernandoRomera:
Some news about this?

simahawk:
> Some news about this?

no news. I have to get back to it. Probably next week, not sure...

FernandoRomera:
Thanks, we are really interested in PR https://github.com/OCA/edi/pull/295 of the EDI repository, since the approach is much simpler (or we see it better) than the current one.

Current:
_usage = "edi.input.validate.demo_backend.test_csv_input"

Better:
 _usage = "input.validate"
_backend_type = "demo_backend"
_exchange_type = "test_csv_input"

component/core.py
simahawk: fixed

component/core.py
simahawk: @guewen should be better now. `component` is left untouched (apart moving out some utils) and `query_component` does the job for the fault tolerant lookup.

component/core.py
```diff
@@ -410,13 +450,27 @@ def component(self, usage=None, model_name=None):
...
+        component_classes, work_context = self._matching_components(
+            usage=usage, model_name=model_name, **kw
+        )
+        if len(component_classes) > 1:
+            component_classes = self._filter_components_by_collection(component_classes)
+        if len(component_classes) > 1:
+            component_classes = self._filter_components_by_model(
+                component_classes, model_name
+            )
+        return component_classes[0](work_context) if component_classes else None
```
guewen: Thanks for the changes. The interrogation I had on the previous version remains. If more than one component match, the result is not deterministic (I assume, maybe you checked this). Then, as a user of the API, I don't know which component will be returned. How can it ever make sense?

simahawk: well, for instance, you may try to lookup a component by a sequence of usage candidates or by a sequence of other parameters. The order in this case does not matter, especially because normally you expect a component to be there or not. If it's not there you try w/ other parameters or simply you don't serve that feature for that given context/model/case.
Doing this w/out having to handle every time a try/except w/ relative import of the exception etc is munch more handy and clean, especially if you want to do this in many places and you don't won't to define your own api/util to handle this in the same way everywhere.
Hope you get my point :)

component/core.py
```diff
@@ -389,16 +433,12 @@ def component(self, usage=None, model_name=None):
         elif len(component_classes) > 1:
             # If we have more than one component, try to find the one
             # specifically linked to the collection...
-            component_classes = [
-                c for c in component_classes if c._collection == self.collection._name
-            ]
+            component_classes = self._filter_components_by_collection(component_classes)
```
simahawk, line 436: @guewen seems to me that this filtering on collection and model it's done twice because `registry.lookup` already does it. Am I wrong?

component/core.py
guewen: > The order in this case does not matter, especially because normally you expect a component to be there or not.

You expect A component.
Ok, but you mention a case where you have one component or zero.

But again, if several match? You say the order does not matter. I don't get why. If the list of matching component is > 1 and you return [0], and there is no way to control what will be the first, you may have a component and after a restart another component, that just feels wrong.

guewen: And I think it IS wrong

simahawk: Well, yes, in theory it would be better to have a way to specify the ordering of the components.
Eg: if I have 2 params to match and I want to give precedence w/ this order:
1. param1 + param2 match
2. param1 only match
3. param2 only match

How would you handle this?
Right now there's no way to do it because there's no sorting and `_component_match` works on each single component class.

What you'll suggest -  I guess - is: use `many_components` sort resulting components and pick what you want. Right?

guewen: > What you'll suggest - I guess - is: use many_components sort resulting components and pick what you want. Right?

Yes, it gives control to the user of the API.

component/core.py
guewen: `registry.lookup` filters on the same collection or no collection. Here, we narrow down to a single component matching the collection if we have one component on the collection and one without.

simahawk:
@guewen as per our discussion: I kept only kw args propagation and split to private methods for reusable code.

guewen (approved):
Thanks!

etobella (approved):

PR #400, opened 2021-04-16
13.0 component: improve README

gurneyalex:

PR #401, opened 2021-04-16
14.0 component: improve README

gurneyalex:


simahawk (approved):
LG. A subtitle as proposed in #400 would be better IMO

marcelsavegnago (approved):

guewen:
Hey, thanks. I would not advertise `TransactionCase` as there isn't any good reason to use it vs `SavepointCase` (I think it is even removed in the core on master)

guewen (approved):

PR #385, opened 2020-10-16
[14.0][MIG] connector

eantones:
Depends on:
- [x] [OCA/queue#263](https://github.com/OCA/queue/pull/263)
- [x] [OCA/connector#382](https://github.com/OCA/connector/pull/382)
- [x] [OCA/connector#384](https://github.com/OCA/connector/pull/384)

PR #403, opened 2021-04-24
[14.0] Migrate test_connector

guewen:
The migration of the test module was missing in https://github.com/OCA/connector/pull/385


guewen:
Tested locally :+1:

PR #404, opened 2021-05-04
[14.0] Migrating improvements

etobella:
Necessary commits for 14.0 `edi` version

Cherry-picked from 13.0

@simahawk

### DIFF
--- a/component/core.py
+++ b/component/core.py
@@ -370,6 +370,33 @@ class WorkContext(object):
                 matching_components.append(cls)
         return matching_components
 
+    def _filter_components_by_collection(self, component_classes):
+        return [c for c in component_classes if c._collection == self.collection._name]
+
+    def _filter_components_by_model(self, component_classes, model_name):
+        return [
+            c
+            for c in component_classes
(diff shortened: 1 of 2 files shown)


component/core.py
```diff
@@ -346,14 +346,58 @@ def component_by_name(self, name, model_name=None):
...
             self.collection._name, usage=usage, model_name=model_name
         )
+        matching_components = []
+        for cls in component_classes:
+            try:
+                matching = cls._component_match(
+                    self, usage=usage, model_name=model_name, **kw
+                )
+            except TypeError as err:
+                # Backward compat
```
simahawk, line 360: I think we can drop this

etobella: I would agree with you if this was the migration of the module, but component was already on 14, could be implementations with the old behavior, isn't it?

simahawk: oh yeah... well, it's unlikely that someone apart from @guewen customized it  but... ok, let's stay safe and do it on v15 :+1:

simahawk (approved):

PR #407, opened 2021-06-23
[13.0][FIX] component: Inherit ComponentRegistryCase from odoo.common.TreeCase

sergio-teruel:
This PR fix missing methods
cc @Tecnativa 
ping @carlosdauden @joao-p-marques 

Before this PR:

```
2021-06-23 14:19:58,419 65 [1;32m[1;49mINFO[0m test odoo.modules.loading: Module component_event: loading demo 
2021-06-23 14:19:58,439 65 [1;32m[1;49mINFO[0m test odoo.modules.module: odoo.addons.component_event.tests.test_event running tests. 
2021-06-23 14:19:58,440 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_collect_several ... 
2021-06-23 14:19:58,496 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_event ... 
2021-06-23 14:19:58,552 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_event_cache ... 
2021-06-23 14:19:58,608 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_event_cache_collection ... 
2021-06-23 14:19:58,664 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_event_cache_model_name ... 
2021-06-23 14:19:58,720 65 [1;32m[1;49mINFO[0m test odoo.addons.component_event.tests.test_event: Starting TestEvent.test_skip_if ... 
2021-06-23 14:19:58,779 65 [1;33m[1;49mWARNING[0m test odoo.modules.loading: Transient module states were reset 
2021-06-23 14:19:58,784 65 [1;31m[1;49mERROR[0m test odoo.modules.registry: Failed to load registry 
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 423, in load_modules
    loaded_modules, update_module, models_to_check)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 315, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 257, in load_module_graph
    report.record_result(odoo.modules.module.run_unit_tests(module_name))
  File "/opt/odoo/custom/src/odoo/odoo/modules/module.py", line 640, in run_unit_tests
    result = OdooTestRunner().run(suite)
  File "/opt/odoo/custom/src/odoo/odoo/modules/module.py", line 609, in run
    test(result)
  File "/usr/local/lib/python3.6/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
  File "/usr/local/lib/python3.6/unittest/suite.py", line 112, in run
    self._tearDownPreviousClass(test, result)
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 251, in _tearDownPreviousClass
    previousClass.doClassCleanups()
AttributeError: type object 'TestEvent' has no attribute 'doClassCleanups'
2021-06-23 14:19:58,785 65 [1;37m[1;41mCRITICAL[0m test odoo.service.server: Failed to initialize database `test`. 
  File "/opt/odoo/custom/src/odoo/odoo/service/server.py", line 1192, in preload_registries
    registry = Registry.new(dbname, update_module=update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/registry.py", line 86, in new
    odoo.modules.load_modules(registry._db, force_demo, status, update_module)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 423, in load_modules
    loaded_modules, update_module, models_to_check)
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 315, in load_marked_modules
    perform_checks=perform_checks, models_to_check=models_to_check
  File "/opt/odoo/custom/src/odoo/odoo/modules/loading.py", line 257, in load_module_graph
    report.record_result(odoo.modules.module.run_unit_tests(module_name))
  File "/opt/odoo/custom/src/odoo/odoo/modules/module.py", line 640, in run_unit_tests
    result = OdooTestRunner().run(suite)
  File "/opt/odoo/custom/src/odoo/odoo/modules/module.py", line 609, in run
    test(result)
  File "/usr/local/lib/python3.6/unittest/suite.py", line 84, in __call__
    return self.run(*args, **kwds)
  File "/usr/local/lib/python3.6/unittest/suite.py", line 112, in run
    self._tearDownPreviousClass(test, result)
  File "/opt/odoo/custom/src/odoo/odoo/tests/common.py", line 251, in _tearDownPreviousClass
    previousClass.doClassCleanups()
AttributeError: type object 'TestEvent' has no attribute 'doClassCleanups'
```

After this PR, the test has been fixed

I don't know if this is the best way to fix the test...

### DIFF
--- a/component/tests/common.py
+++ b/component/tests/common.py
@@ -2,7 +2,6 @@
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
 
 import copy
-import unittest
 from contextlib import contextmanager
 
 import odoo
@@ -105,7 +104,7 @@ class SavepointComponentCase(common.SavepointCase, ComponentMixin):
 
 
 class ComponentRegistryCase(
-    unittest.TestCase, common.MetaCase("DummyCase", (object,), {})
+    common.TreeCase, common.MetaCase("DummyCase", (object,), {})
 ):
     """ This test case can be used as a base for writings tests on components
 
--- a/component_event/tests/test_event.py
+++ b/component_event/tests/test_event.py
@@ -5,7 +5,7 @@ import unittest
 
 import mock
 
-from odoo.tests.common import tagged
+from odoo.tests.common import tagged, TreeCase
 
 from odoo.addons.component.core import Component
 from odoo.addons.component.tests.common import (
@@ -17,7 +17,7 @@ from odoo.addons.component_event.core import EventWorkContext
 
 
 @tagged("standard", "at_install")
-class TestEventWorkContext(unittest.TestCase):
+class TestEventWorkContext(TreeCase):
     """ Test Events Components """
 
     def setUp(self):


component_event/tests/test_event.py
```diff
@@ -17,7 +17,7 @@
 
 
 @tagged("standard", "at_install")
-class TestEventWorkContext(unittest.TestCase):
+class TestEventWorkContext(TreeCase):
```
joao-p-marques, line 20: I think it would be better to inherit from `TransactionCase` or `SavepointCase` here, if possible, to standardize the behavior... 

It would also eliminate the need of the explicit tag before, I think...

component/tests/common.py
```diff
@@ -105,7 +104,7 @@ def setUp(self):
 
 
 class ComponentRegistryCase(
-    unittest.TestCase, common.MetaCase("DummyCase", (object,), {})
+    common.TreeCase, common.MetaCase("DummyCase", (object,), {})
```
joao-p-marques, line 107: With the other change mentioned, I think this should be left as is... The docstring in this class explains that this is being created for a specific purpose, so maybe it is best to leave it, even when Odoo's `TreeCase` inherits from `unittest.TestCase`...

A better solution would be not to use this class downstream where we don't need, but instead use one of Odoo's `common` cases.

joao-p-marques (changes requested):
@sergio-teruel maybe this suggestions could help without changing too much and causing other problems... Although I am also not sure why this different test behavior is needed here. Maybe @simahawk can give us a clue?

Also, please check Travis, there seem to be some pre-commit errors...

component/tests/common.py
simahawk: this class should not inherit from any odoo class. It's a base class for tests class like `TransactionComponentRegistryCase` or `SavepointComponentRegistryCase` which already inherit from `TreeCase`

component_event/tests/test_event.py
simahawk: this is wrong as well. This is a unit test, not an odoo related test.

simahawk (changes requested):
would be nice to understand the use case for this change. 
Is there any PR failing?

simahawk:
it seems more an error due to python version because that method is here https://docs.python.org/3.8/library/unittest.html#unittest.TestCase.doClassCleanups for v >= 3.8. 
So, I wonder: how comes that it was working before. With the TreeCase they support versions below 3.8 indeed :thinking:

pedrobaeza:
So are you using something of Python 3.8 only?

sergio-teruel:
Yes @simahawk You are right...
TreeCase implents doClassCleanups here (https://github.com/odoo/odoo/blob/c6c8c84a6bf3c99bb4d7d80263a41782f058acb2/odoo/tests/common.py#L263). That is the reason why I inherit from TreeCase...

We use python 3.6 and component module tests are failing due to this method has not exists...
Anyway, what is the correct way to correct this?
Thanks

simahawk:
> So are you using something of Python 3.8 only?

that's the point: I'm not using py3.8 but I've never seen this test fail.
Anyway, I'll try to run test locally from scratch. 
Still, I'd like to understand how comes that the last build from 19d ago was green if nothing has changed.

simahawk:
ohh, now I see: they've screwed bare unit tests. The odoo test suite machinery assumes that method is always there :sob: 
With pytest it works. 

FTR the support for this func has been added to pytest as well in v6.2.0 (see https://github.com/pytest-dev/pytest/pull/8033) BUT is called only if available, for backward compat.

Please check https://github.com/OCA/connector/pull/408 which works w/ both pytest and odoo suite and it's a smaller conditional change :pray:

simahawk:
merging #408

PR #408, opened 2021-06-24
component_event: fix unit test w/ odoo test suite

simahawk:
This commit
https://github.com/odoo/odoo/commit/ed831abd7d91ee873a1a92ecea2de316a025754f
introduced a backward imcompatible change
that makes simple unit tests ran through Odoo suite fail
because `doClassCleanups` is not defined.
They defined it on `TreeCase` so... there we go.

CC @sergio-teruel

### DIFF
--- a/component_event/tests/test_event.py
+++ b/component_event/tests/test_event.py
@@ -1,6 +1,7 @@
 # Copyright 2017 Camptocamp SA
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
 
+import sys
 import unittest
 
 import mock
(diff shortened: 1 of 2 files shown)


sergio-teruel (approved):

joao-p-marques (approved):
LGTM :+1: 

Just out of curiosity, why not to always inherit from `TreeCase` then? It could avoid similar problems in the future, I guess...

sergio-teruel:
Tested locally ok.... Travis is failing

carlosdauden (approved):

simahawk:
> LGTM +1
> 
> Just out of curiosity, why not to always inherit from `TreeCase` then? It could avoid similar problems in the future, I guess...

Mainly 3 reasons:

1. make clear why we need this
2. make clear that we can trash it in future versions
3. make sure we can still test w/ bare unittest classes (there's no real reason to depend on odoo specific stuff)

sergio-teruel:
Thanks @simahawk

joao-p-marques:
> > LGTM +1
> > Just out of curiosity, why not to always inherit from `TreeCase` then? It could avoid similar problems in the future, I guess...
> 
> Mainly 3 reasons:
> 
>     1. make clear why we need this
> 
>     2. make clear that we can trash it in future versions
> 
>     3. make sure we can still test w/ bare unittest classes (there's no real reason to depend on odoo specific stuff)

Ok, thanks :+1: 🙂

Issue #415, opened 2021-09-08
Unittests Fail with Dependant Modules to be tested


agent-z28:
We have two modules to be tested:
test_1 and test_1_addon(depends test_1).

now we run the unittests for the module with  

odoo -i test_1,test_1_addon --stop-after-init --workers=0 --db-filter=devel --test-enable.

then the registry will be build here:

https://github.com/OCA/connector/blob/14.0/component/tests/common.py#L196

current_addon = test_1 

and exclude our first addon here:

https://github.com/OCA/connector/blob/faa50cba80f80d275a30ea000f2446105909f7a3/component/builder.py#L70

in the module_list we have e.g. ['bus', 'web_editor', 'web', 'base_setup', 'web_tour', '**test_1_addon**', 'base_edi', 'queue_job', 'fetchmail', 'auth_signup', 'web_settings_dashboard', 'mail', 'web_planner', 'web_kanban_gauge', 'web_diagram', 'component_event', 'base_import', 'component', 'base_sparse_field', 'auth_crypt'] 
exclude_addon ['test_1']

2021-09-08 08:27:10,894 1 ERROR devel odoo.modules.graph: module test_1_addon: Unmet dependencies: test_1

The comment says:

        # build the components of every installed addons already installed
        # but the current addon (when running with pytest/nosetest, we
        # simulate the --test-enable behavior by excluding the current addon
        # which is in 'to install' / 'to upgrade' with --test-enable).

But actually, it runns better without the exclude_addon logic.

What was the idea for this code?

agent-z28:
@guewen what you say

guewen:
If you don't want this behavior, you should use the `SavepointComponentCase`. 
The `ComponentRegistryCase` excludes the current addon components so you can load **only** the component you want to test.
It is widely used in the core tests for instance.
But for an implementation, you may not need it at all.

agent-z28:
@guewen i will check it this week

simahawk:
closing for inactivity, feel free to reopen

PR #416, opened 2021-09-08
imp(component): remove exclude_addon logic for unittests to avoid dep…

agent-z28:
fix https://github.com/OCA/connector/issues/415


guewen:
Can you check my anwer on https://github.com/OCA/connector/issues/415#issuecomment-915104476?

PR #419, opened 2021-10-25
[IMP] connector: add missing exception

chandni299:
[IMP] connector: add missing exception


gurneyalex:
hello @chandni299 can you give some context on why you feel this change is needed? Adding an exception class definition but not raising that exception anywhere feels strange.

PR #420, opened 2021-11-19
[15.0][MIG] component

ajaniszewska-dev:


component/i18n/am.po
```diff
@@ -0,0 +1,47 @@
+# Translation of Odoo Server.
```
simahawk: these translation files can be trashed IMO, not mandatory of course :)

component/tests/common.py
```diff
@@ -0,0 +1,264 @@
...
+        # super)
+        common.TransactionCase.setUp(self)
+        ComponentMixin.setUp(self)
+        # There's no env on setUpClass of TransactionCase, must do it here.
+        self.env.context = dict(
+            self.env.context, components_registry=self._components_registry
+        )
+
+
+class SavepointComponentCase(common.SavepointCase, ComponentMixin):
```
simahawk: This class should be deprecated as common.SavepointCase.
I would make it inherit from `TransactionComponentCase` directly and just log deprecation.
Setup code should be moved from setUp to setUpClass.

component/tests/common.py
```diff
@@ -0,0 +1,264 @@
...
+        common.TransactionCase.setUp(self)
+        ComponentRegistryCase._setup_registry(self)
+        self.collection = self.env["collection.base"]
+
+    def tearDown(self):
+        ComponentRegistryCase._teardown_registry(self)
+        common.TransactionCase.tearDown(self)
+
+
+class SavepointComponentRegistryCase(common.SavepointCase, ComponentRegistryCase):
```
simahawk: same

component/tests/common.py
```diff
@@ -0,0 +1,264 @@
...
+    Or::
+
+        collection_record = self.env['my.collection'].browse(1)
+        with collection_record.work_on(
+                'res.partner',
+                components_registry=self.comp_registry) as work:
+
+    """
+
+    if sys.version_info < (3, 8):
```
simahawk: this should be trashed if we don't run on 3.7 anymore

simahawk (changes requested):

simahawk:
pls check pre-commit too

component/tests/test_build_component.py
```diff
@@ -0,0 +1,274 @@
...
+
+        Component1._build_component(self.comp_registry)
+        msg = "Component.*already exists.*"
+        with self.assertRaisesRegex(TypeError, msg):
+            Component2._build_component(self.comp_registry)
+
+    def test_no_parent(self):
+        """Ensure we can't _inherit a non-existent component"""
+
+        class Component1(Component):
```
ajaniszewska-dev: `component/tests/test_build_component.py:59: [R7980(consider-merging-classes-inherited), TestBuildComponent.test_inherit_bases.Component2] Consider merging classes inherited to "component1" from test_build_component.py:62, test_build_component.py:184, test_build_component.py:269.
`

@simahawk  pre commit complains for that as above. Possible to disable it somehow?

simahawk: in this case just customize the class w/ a silly method or attribute that is never used, eg:

```
class Component1(Component):
    def say_hello(self):
        """Dumb method."""
        return "hello"
```

ajaniszewska-dev: @simahawk  thanks, but I tried both this methods and params, to me it doesn't work.
`# pylint: disable=R7980` above affected classes maybe?

simahawk: Checked. Just do this: 
1. rollback your changes to address this
2. add this block after the file header

```py
# Copyright 2017 Camptocamp SA
# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)

# Tell pylint to not bother us for all our fake component classes
# pylint: disable=consider-merging-classes-inherited
```

simahawk:
@ajaniszewska-dev if you delete commits w/ push force, the comment done on specific commit lose reference. Just use fixup for now :pray:

simahawk:
> @ajaniszewska-dev if you delete commits w/ push force, the comment done on specific commit lose reference. Just use fixup for now pray

or maybe is github is just crazy here because is the 1st time I see this messed up summary of commits :thinking:

simahawk:
@ajaniszewska-dev there are still things to be fixed. Could you please propose a PR to add the module as not installable first? We can fast track it and then we'll have an easier review here.

lmignon:
To ease the migration and avoid all this mess with the commit histrory, I propose to recreate the 15.0 branch from the 14.0 with all the addons declared with `"installable": False. What do you think about @simahawk @guewen @yankinmax

simahawk:
@lmignon I was thinking the same but I had some issues w/ my MT setup and I had no time yesterday.
I'll see if I can do it today, unless any of you can make it before :wink:

lmignon:
@simahawk I'll do it into the coming hours

lmignon:
@yankinmax @simahawk @guewen 15.0 is now initialized from 14.0...

dreispt:
@liweijie0812 @simahawk @guewen I updated https://github.com/OCA/connector/issues/418 accordingly.

PR #423, opened 2021-11-25
[15.0][MIG] component: Migration to 15.0

yankinmax:
ref: BSRD-470


component/tests/common.py
```diff
@@ -152,75 +132,57 @@ class ComponentRegistryCase(
...
-                function, args, kwargs = cls._class_cleanups.pop()
-                try:
-                    function(*args, **kwargs)
-                except Exception:
-                    cls.tearDown_exceptions.append(sys.exc_info())
-
-    @staticmethod
-    def _setup_registry(class_or_instance):
+    def setUp(self):
+        super(ComponentRegistryCase, self).setUp()
```
simahawk: this is wrong, please keep it as before.
This must run on `setUpClass` but keep it in a separate method.

component/tests/common.py
```diff
@@ -152,75 +132,57 @@ class ComponentRegistryCase(
...
-                components_registry=class_or_instance.comp_registry,
+            self.env.context = dict(
+                self.env.context,
+                components_registry=self.comp_registry,
             )
 
-    @staticmethod
-    def _teardown_registry(class_or_instance):
+    def tearDown(self):
+        super().tearDown()
```
simahawk: same

simahawk (changes requested):

yankinmax:
@simahawk it used to work. Now registry isn't cleaned anymore. I got errors on tests:
```
TypeError: Component 'component1' (in class <class 'odoo.addons.component.tests.test_build_component.TestBuildComponent.test_register.<locals>.Component1'>) already exists
```
Tried `self.registry.reset_changes()` from
https://www.odoo.com/documentation/15.0/developer/reference/backend/testing.html#odoo.tests.common.TransactionCase.
Doesn't help.

yankinmax:
@simahawk it runs without issues locally, and also `base_rest` with this variant runs

simahawk:
the point is not if it works or not. The point is that the setup should happen on the class and not on each test. I'll have a look tomorrow.

yankinmax:
> the point is not if it works or not. The point is that the setup should happen on the class and not on each test. I'll have a look tomorrow.

I hope it's. I watched your OCA video about testing techniques

component/tests/common.py
```diff
@@ -152,75 +132,65 @@ class ComponentRegistryCase(
...
+        super().setUp()
+        ComponentRegistryCase._setup_registry(self)
+
+    def tearDown(self):
+        super().tearDown()
+        ComponentRegistryCase._teardown_registry(self)
 
     @staticmethod
-    def _setup_registry(class_or_instance):
+    def _setup_registry(self):
```
simahawk: this is staticmethod, shouldn't have `self`

component/tests/common.py
```diff
@@ -152,75 +132,65 @@ class ComponentRegistryCase(
...
         # it builds the 'final component' for every component of the
         # 'component' addon and push them in the component registry
-        class_or_instance.comp_registry.load_components("component")
+        self.comp_registry.load_components("component")
         # build the components of every installed addons already installed
         # but the current addon (when running with pytest/nosetest, we
         # simulate the --test-enable behavior by excluding the current addon
         # which is in 'to install' / 'to upgrade' with --test-enable).
-        current_addon = _get_addon_name(class_or_instance.__module__)
-        with new_rollbacked_env() as env:
```
simahawk: I don't get why you refactored this part. Is not doing the same thing :thinking:

component/tests/common.py
```diff
@@ -83,30 +86,8 @@ def setUp(self):
...
-    def setUp(self):
-        # resolve an inheritance issue (common.SavepointCase does not call
-        # super)
-        common.SavepointCase.setUp(self)
-        ComponentMixin.setUp(self)
-
-
 class ComponentRegistryCase(
-    unittest.TestCase, common.MetaCase("DummyCase", (object,), {})
+    common.BaseCase, common.MetaCase("DummyCase", (object,), {})
```
simahawk: There's a reason why it has been done w/ the basic unittest class.

simahawk (changes requested):

component/tests/common.py
yankinmax: you told to leave as it's, I left.

component/tests/common.py
yankinmax: it seems to me typical among modules I've seen during this week, so I just choose the most recent one

component/tests/common.py
```diff
@@ -152,31 +128,13 @@ class ComponentRegistryCase(
...
-            cls.tearDown_exceptions = []
-            while cls._class_cleanups:
-                function, args, kwargs = cls._class_cleanups.pop()
-                try:
-                    function(*args, **kwargs)
-                except Exception:
-                    cls.tearDown_exceptions.append(sys.exc_info())
+    def setUp(self):
+        super().setUp()
+        ComponentRegistryCase._setup_registry(self)
```
simahawk: Again, this goes to `setUpClass`.

yankinmax: So, then I can't pass tests locally.
Here what I have:
```
...
@classmethod
    def setUpClass(cls):
        super().setUpClass()
        ComponentRegistryCase._setup_registry(cls)

    @classmethod
    def tearDownClass(cls):
        super().tearDownClass()
        ComponentRegistryCase._teardown_registry(cls)

    @staticmethod
    def _setup_registry(class_or_instance):
...
```
Then if I run tests I got this:
```
if cls._name in registry and not parents:
>           raise TypeError(
                "Component %r (in class %r) already exists. "
                "Consider using _inherit instead of _name "
                "or using a different _name." % (cls._name, cls)
            )
E           TypeError: Component 'component1' (in class <class 'odoo.addons.component.tests.test_build_component.TestBuildComponent.test_check_parent_abstract_over_component.<locals>.Component1'>) already exists. Consider using _inherit instead of _name or using a different _name.

odoo/external-src/connector/component/core.py:830: TypeError
```
I'm not an expert but think CleanUp wasn't called on set up test class.
I'm not an expert in that kind of stuff, so I try to add what I met in such code:
```
@class_or_instance.addCleanup
        def _close_and_roolback():
            env.cr.rollback()  # we shouldn't have to commit anything
            env.cr.close()
```
And this gives me:
```
@class_or_instance.addCleanup
>       def _close_and_roolback():
E       TypeError: addCleanup() missing 1 required positional argument: 'function'
```
Can you help me pls?

simahawk: _close_and_rollback is not here anymore as I asked you to rollback your changes and keep the old implementation. Have you tried again after that?

dreispt (changes requested):
Missing commit history
Please follow the migration procedure here:
https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

simahawk:
> Missing commit history Please follow the migration procedure here: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0

@dreispt you are late :rofl: The history is there because the repo has been re-initialized w/ modules uninstallable :wink: 

See https://github.com/OCA/connector/pull/420#issuecomment-978917294

component/tests/common.py
```diff
@@ -83,28 +82,6 @@ def setUp(self):
         )
 
 
-class SavepointComponentCase(common.SavepointCase, ComponentMixin):
```
simahawk, line 86: @yankinmax this is still needed but should inherit from `common.TransactionCase` and be renamed to `ComponentCase` or `TransactionComponentCase`.
Is not the same one as `TransactionComponentRegistryCase`.

simahawk (changes requested):

simahawk:
pls squash the last commit once you have done

component/tests/common.py
yankinmax: I left TransactionComponentCase and removed SavepointComponentCase. Just open code of commit pls.

simahawk: I'm looking at the full diff. There's no class named `TransactionComponentCase` and should stay here.
As said, this is not the same as `TransactionComponentRegistryCase`.

yankinmax: -https://github.com/yankinmax/connector/blob/370efd265e3aa5637aa5918da3b6fb9ce99cd913/component/tests/common.py#L59

dreispt:
@simahawk I was just flying by. Not that I read it, but I updated https://github.com/OCA/connector/issues/418 accordingly.

component/tests/common.py
simahawk: ah yeah, sorry, the expanding of the diff tricked me. That class is already good as is.

simahawk (approved):

simahawk:
> @simahawk I was just flying by. Not that I read it, but I updated #418 accordingly.

alright, thanks!

simahawk (changes requested):
I've just realized you miss https://github.com/OCA/connector/blob/15.0/test_component

yankinmax:
@simahawk definitely didn't pay attention to other modules. Will also update test_component, but to me makes sense not to open separate PR and add just new commit here. Is it ok?

simahawk:
no, because the module is called `test_component` for a reason ;)

simahawk:
test_component migrated on https://github.com/OCA/connector/pull/424

component/tests/common.py
```diff
@@ -231,34 +181,41 @@ def _build_components(self, *classes):
...
-    # pylint: disable=W8106
-    def setUp(self):
-        # resolve an inheritance issue (common.TransactionCase does not use
-        # super)
-        common.TransactionCase.setUp(self)
-        ComponentRegistryCase._setup_registry(self)
-        self.collection = self.env["collection.base"]
+    This class doesn't set up the registry for you.
+    You're supposed to explicitly call `_setup_registry` and `_teardown_registry`
+    when you need it, either on setUpClass and tearDownClass or setUp and tearDown.
```
sebalix, line 188: Is it supposed to work like that?
In previous versions the registry was set up automatically when using this test class, and now every tests depending on that should explicitly call these methods?
I don't know how it works actually but that seems strange from a user POV. The docstring of `ComponentRegistryCase` says that I could use this class if I want to get the full control of the registry.

simahawk: it will be like this. We don't have/need 2 classes anymore and I find better that each implementation clearly shows where the setup is done. If you want to "hide" or replicate it you define your own common class.
`ComponentRegistryCase` is another case where you register your own custom test components.

sebalix: @simahawk thanks, so it has been discussed :)

sebalix (approved):

simahawk (approved):

john-herholz-dt:
Hello @simahawk , can you give a short explanation why SavepointComponentCase was removed from the tests?
I work with a module that relies on it.

Thanks

danielduqma:
Hello @john-herholz-dt,

I just ran into the same problem, and I found that `SavepointCase` was merged into `TransactionCase`. Just replace `SavepointComponentCase` by `TransactionComponentCase` and it should work. [More info](https://github.com/odoo/odoo/pull/62031).

Regards.

PR #424, opened 2021-11-26
[15.0][MIG] test_component: Migration to 15.0

yankinmax:
Depends on:
- [ ] https://github.com/OCA/connector/pull/423


test-requirements.txt
```diff
@@ -0,0 +1 @@
+odoo-addon-component @ git+https://github.com/OCA/connector@refs/pull/423/head#subdirectory=setup/component
```
simahawk: next time put this in its own commit to make it prominent

simahawk (commented):
LG, I wait for one more review on #423

test-requirements.txt
yankinmax: ok, got it, thanks

simahawk (changes requested):
blocking till the test dep is gone

sebalix (approved):
Pre-approving

simahawk (approved):

PR #425, opened 2021-12-02
[MIG][15.0] connector + component_event + test_connector

i-vyshnevska:


simahawk:
`test_connector` should be migrated too

component_event/tests/test_event.py
```diff
@@ -47,6 +47,11 @@ def doClassCleanups(cls):
                 except Exception:
                     cls.tearDown_exceptions.append(sys.exc_info())
```
simahawk: the block above should be removed

simahawk (changes requested):

i-vyshnevska:
@simahawk ready for review

connector/components/synchronizer.py
```diff
@@ -55,7 +56,7 @@ def run(self):
     def mapper(self):
         """Return an instance of ``Mapper`` for the synchronization.
 
-        The instanciation is delayed because some synchronisations do
+        The instantiations is delayed because some synchronizations do
```
simahawk:
```suggestion
        The instantiation is delayed because some synchronizations do
```

yankinmax: as you're fixing mistakes, you can also choose between `instantiations are` and `instantiation is`

component_event/tests/test_event.py
```diff
@@ -162,7 +139,7 @@ class TestEvent(ComponentRegistryCase):
 
     def setUp(self):
         super(TestEvent, self).setUp()
```
yankinmax, line 141:
```suggestion
        super().setUp()
```

component_event/tests/test_event.py
```diff
@@ -392,6 +369,7 @@ class TestEventFromModel(TransactionComponentRegistryCase):
 
     def setUp(self):
         super(TestEventFromModel, self).setUp()
```
yankinmax, line 371:
```suggestion
        super().setUp()
```

component_event/tests/test_event.py
```diff
@@ -21,31 +20,9 @@
...
-            cls.tearDown_exceptions = []
-            while cls._class_cleanups:
-                function, args, kwargs = cls._class_cleanups.pop()
-                try:
-                    function(*args, **kwargs)
-                except Exception:
-                    cls.tearDown_exceptions.append(sys.exc_info())
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.test_sequence = 0
```
yankinmax, line 25: I didn't meet such code before. Could you pls explain what `self.test_sequence = 0` is for?

saxomanu:
Is it ready to merge ? what is missing ? Can i help ?

i-vyshnevska:
@simahawk can you pass another review

component_event/tests/test_event.py
simahawk: @i-vyshnevska ?

simahawk:
@i-vyshnevska pls squash when done

component_event/tests/test_event.py
i-vyshnevska: the field is defined in odoo metatest class
 https://github.com/odoo/odoo/blob/5cfce06bf92cc746b4dc3b2b554ad57daac1af1d/odoo/tests/common.py#L296
 but us we inherit here form uittetst we also need this field so odoo will be able to run test

gurneyalex (approved):

component_event/tests/test_event.py
yankinmax: @i-vyshnevska thanks for the explanation

simahawk (approved):

simahawk (changes requested):
@i-vyshnevska there's a mistake. 
There's a commit to migrate "test_connector" but I don't see "test_component".
Also that commit contains unrelated changes https://github.com/OCA/connector/pull/425/commits/1e0c578cd1bbc2cd50ed8225518deaa046710010
We need test_component here. Plus, I wonder how tests for test_connector

simahawk (approved):
Ah, my bad, I got confused by names. Re-checked 15.0, LG.

PR #432, opened 2022-05-05
[14.0][FIX] Get rid of DeprecationWarning in Mapper

florentx:
Having these warnings in log files:
```
WARNING odoodb py.warnings: /odoo/external-src/connector/connector/components/mapper.py:806:
   DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated,
   and in 3.8 it will stop working

|   if isinstance(from_attr, collections.Callable):
|  
```

### DIFF
--- a/connector/components/mapper.py
+++ b/connector/components/mapper.py
@@ -11,10 +11,10 @@ external records into Odoo records and conversely.
 
 """
 
-import collections
 import logging
 from collections import namedtuple
 from contextlib import contextmanager
+from typing import Callable
 
 from odoo import models
 
@@ -127,7 +127,7 @@ def none(field):
     """
 
     def modifier(self, record, to_attr):
-        if isinstance(field, collections.Callable):
+        if isinstance(field, Callable):
             result = field(self, record, to_attr)
         else:
             result = record[field]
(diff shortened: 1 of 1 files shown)


simahawk:
Thanks! Can you backport this instead https://github.com/OCA/connector/pull/431 ? I had no time yet to do it.

florentx:
Why not used `callable()`? it is as reliable as the ABC alternative, and easier to understand. Maybe slightly faster as it is builtin.

simahawk:
TBH I cannot find a specific reference to any guideline or best practice for this. Do you have any?

Back to py2 I used always `callable` but then it got somehow deprecated - at least from a best practice POV - as you can see here :)

Maybe @guewen as a better picture. Maybe is just a religion battle :smile:

florentx:
I can explain, as I was involved in Py core dev at this time:

- in Python 2 it was `callable()`
- in Python 3 revamp they decided to drop it (reco was `isinstance(ob, collections.Callable)`)
- but on 2nd though (and complaints) they put it back in Python 3.3, and it is the recommended syntax since

For OCA connector, it was converted from `callable()` to `isinstance(..., Callable)` by the tool `2to3` automatically (old version of the tool).

But we should revert it to the `callable()` I believe, because nobody needs to support versions 3.1 and 3.2 of Python.

guewen (approved):
Thanks!

simahawk:
Cool, thanks for the explanation!
Still, I'd recommend to cherry-pick the commit from #431 first to not have false positives when back/forward porting commits across branches.
Could you do that pls? Then we can fwd port yours when this PR is merged. Thank you in advance! :pray: 
Or let me know and I'll do it later.

PR #442, opened 2022-10-04
[16.0][MIG] connector + test_connector: Migration to 16.0

asierneiradev:
#440 

Dependencies : 

- [x] [queue_job](https://github.com/OCA/queue/tree/16.0/queue_job)
- [x] [component](https://github.com/OCA/connector/tree/16.0/component)
- [x] [component_event](https://github.com/OCA/connector/pull/443)

PR #447, opened 2022-10-10
[16.0]connector: Migration 16.0

bizzappdev:


simahawk:
There's already a migration here https://github.com/OCA/connector/pull/442

PR #457, opened 2022-12-28
[13.0][FIX] component_event: handle empty vals on create.

LoisRForgeFlow:
Upstream `create` method does support an empty vals list. In such case no error is raised, instead an empty recordset is returned and no record is created.

This is useful when using "prepare vals" hook methods, in which you can return `False` if you don't want to create a record. Any extension to the Base model should respect that.

@ForgeFlow


LoisRForgeFlow:
I'm opening the v14 version of this change to use a updated CI config.

PR #458, opened 2022-12-28
[14.0][FIX] component_event: handle empty vals on create.

LoisRForgeFlow:
Upstream `create` method does support an empty vals list. In such case no error is raised, instead an empty recordset is returned and no record is created.

This is useful when using "prepare vals" hook methods, in which you can return `False` if you don't want to create a record. Any extension to the Base model should respect that.

Forward port of https://github.com/OCA/connector/pull/457.


etobella:
If you send an empty vals list `[]` it should work propery :thinking:

PR #460, opened 2023-01-30
[FIX] Update copier template

asierneiradev:
To avoid the isort problem (https://github.com/PyCQA/isort/issues/2077)


asierneiradev:
@guewen @sbidoul ,

Can someone check this and merge it?

Thank you very much

PR #472, opened 2023-10-30
[15.0] component: fix mock import

sonhd91:


guewen:
Hi @sonhd91 thanks for the fix!
You fixed the import correctly, however we should handle `sphinx_bootstrap_theme` differently IMO, I suggest to remove it from this PR and handle it in https://github.com/OCA/connector/pull/471#issuecomment-1785145561

sonhd91:
> Hi @sonhd91 thanks for the fix! You fixed the import correctly, however we should handle `sphinx_bootstrap_theme` differently IMO, I suggest to remove it from this PR and handle it in [#471 (comment)](https://github.com/OCA/connector/pull/471#issuecomment-1785145561)

Okay, branch updated

guewen (approved):
Thanks!

PR #474, opened 2023-11-24
[17.0][MIG] component + test_component: Migration to 17.0

nguyenminhchien:
Ref: BSRD-701


gurneyalex (approved):

component/core.py
```diff
@@ -324,8 +324,9 @@ def component_by_name(self, name, model_name=None):
             and self.collection._name != component_class._collection
         ):
             raise NoComponentError(
-                "Component with name '%s' can't be used for collection '%s'."
-                % (name, self.collection._name)
+                "Component with name '{}' can't be used for collection '{}'.".format(
```
simahawk, line 327: why not fstring as in the other places?

component/core.py
```diff
@@ -0,0 +1,937 @@
...
+            component_classes = self._filter_components_by_collection(component_classes)
+        if len(component_classes) > 1:
+            # ... or try to find the one specifically linked to the model
+            component_classes = self._filter_components_by_model(
+                component_classes, model_name
+            )
+        if len(component_classes) != 1:
+            raise SeveralComponentError(
+                "Several components found for collection '{}', "
+                "usage '{}', model_name '{}'. Found: {}".format(
```
simahawk, line 449: same

simahawk (approved):
minor remarks

component/core.py
nguyenminhchien: it's more convenient to replace `% (name, self.collection._name)` by `.format(name, self.collection._name)`. anw, i have updated the branch to use `fstring` instead.

PR #481, opened 2024-05-01
[16.0] dotfiles update needs manual intervention

sbidoul:
Dear maintainer,

After updating the dotfiles, `pre-commit run -a`
fails in a manner that cannot be resolved automatically.

Can you please have a look, fix and merge?

Thanks,


pedrobaeza:
This one should be handled by one of the PSCs, as the given errors are out of scope of the systematic fixes I'm doing.

PR #498, opened 2025-01-07
[18.0][MIG] connector_base_product: Migration to 18.0

yankinmax:

PR #501, opened 2025-02-04
[18.0][FIX] connector_base_product: remove wrong invisible attribute

yankinmax:
The migration PR was merged with `invisible=True` on the `Connectors` page.
- [ ] https://github.com/OCA/connector/pull/498

I missed that during migration.
The question is: why has it been invisible since v14?
Is anyone (except our customers) using this module?
It doesn't make any sense.
https://github.com/OCA/connector/blob/14.0/connector_base_product/views/product_view.xml


yankinmax:
I have a suggestion to check:
"The page is invisible because there is nothing to display on that page by this module, which only exists to create the notebook page. The modules which want to display something in the notebook page must make the page visible."

PR #504, opened 2025-07-22
[18.0] component: allow ctx override at init

simahawk:
Before this change, the only way to set something in the ctx is to override the env property.
This means that the code would be executed N times and a new env initialized N times.

The original env is preserved and in any case is available on work.env and collection.env.


simahawk:
@guewen mind to have a look at this? :)

guewen (approved):

PR #508, opened 2025-10-11
[19.0][MIG] component

Sanazzzmi:
In Odoo 19, the context has become read-only, so we had to modify the tests to handle the read-only behavior of the context.


fardinmardani (approved):
I tested this code and it works correctly — I ran it myself.

AlphaKamyar (approved):
code tested

adelghaedi (approved):
The code has been tested and works correctly.
Thank you!

Sanazzzmi:
@OCA-git-bot 
@guewen
Please confirm. Thank you.

Sanazzzmi:
@OCA-git-bot
@guewen

Please confirm so we can update other modules.

oerp-odoo:
any update on this? Would be nice to merge it.

component/tests/common.py
```diff
@@ -0,0 +1,214 @@
...
+            # ensure that we load only the components of the 'installed'
+            # modules, not 'to install', which means we load only the
+            # dependencies of the tested addons, not the siblings or
+            # children addons
+            builder.build_registry(comp_registry, states=("installed",))
+            # build the components of the current tested addon
+            current_addon = _get_addon_name(cls.__module__)
+            env["component.builder"].load_components(current_addon)
+            if hasattr(cls, "env"):
+                cls.env.context = dict(
```
mkoeck, line 43: I think here the context is still replaced in the old way which will fail, because the context is readonly

component/tests/common.py
```diff
@@ -0,0 +1,214 @@
...
+        cls.setUpComponent()
+
+    # pylint: disable=W8106
+    def setUp(self):
+        # resolve an inheritance issue (common.TransactionCase does not call
+        # super)
+        common.TransactionCase.setUp(self)
+        ComponentMixin.setUp(self)
+        # There's no env on setUpClass of TransactionCase, must do it here.
+        self.env.context = dict(
```
mkoeck, line 79: I think here the context is still replaced in the old way which will fail, because the context is readonly

mkoeck (changes requested):
Hi, I think there are two more places, where the context is replaced in the old way which will fail due to the context being read only.

component/tests/common.py
mlaitinen: I can confirm this. This causes the tests to fail.

vvrossem:
Hello @Sanazzzmi, thank your for your contribution.
Could you please check the CI error ?

imlopes:
Hello @Sanazzzmi , thank your for this PR.

Could you please check/fix the CI error ?

Thanks !

mostafabarmshory:
Hi @Sanazzzmi,

This seems like a solid approach. Implementing a singleton component module is a great idea. Could you please proceed with the requested changes?

component/readme/CREDITS.md
```diff
@@ -1 +1 @@
-The migration of this module from 17.0 to 18.0 was financially supported by Camptocamp.
+The migration of this module from 18.0 to 19.0 was financially supported by Camptocamp.
```
simahawk, line 1: why this change?

component/tests/common.py
simahawk: last bit to fix

simahawk (changes requested):
LG overall.

simahawk:
replaced by https://github.com/OCA/connector/pull/525

PR #511, opened 2025-11-26
[19.0][MIG] connector: Migration to 19.0

mkoeck:
### Main changes

- **New security model support**
  - Add a `res.groups.privilege` record introduced in 19.0 and adapted the existing res.groups record accordingly

- **Test suite updates for 19.0 API**
  - Replace deprecated `self.env2.reset()` calls with `self.env2.transaction.reset()` in tests.
  - Adapt tests to the now read-only `env.context`:
    - Remove direct writes to `self.env.context`.
    - Use `self.env()` to create a new environment instead
  - Remove / replace references to demo data:
    - Demo loading order changed in 19.0, so some demo `external_id`s are no longer guaranteed to exist.
    - Tests now rely on explicit test data instead of demo records.

### Functional impact

- No functional feature changes are intended.
- The goal of this PR is to make the module compatible with Odoo 19.0 while preserving existing behavior where possible.

### Dependencies

This PR has been tested against the following non-merged branches of dependent modules:

- https://github.com/OCA/connector/pull/525
- https://github.com/OCA/connector/pull/526


vvrossem:
Hello @mkoeck , thank you for your contribution. Could you please attend the CI errors? :pray:

mkoeck:
Hi @vvrossem if I remember correctly the tests are only failing because the dependant modules are not yet updated. I pulled in the open PRs for those locally and I think tests passed. I since learned that it is possible to add a temporary dependency  (or whatever its called), but I am not too familiar with it. If I have the time I will look into it, but these problems should resolve once the PRs for the other modules are merged.

imlopes:
Hello @mkoeck ,
Thank you for this work.

Could you check the failing tests and fix the `pre-commit` too ?
To add the dependency module (while waiting the PR get merged), you can check this commit:
https://github.com/OCA/report-print-send/commit/d87a1d20ca95a3f71051c056e294aaf306a4ba39

mkoeck:
@imlopes I think this looks right now

imlopes:
@mkoeck thank you for this work.
Not sure if it's mandatory (I think not) but it could be better to split your `[MIG]` commit in 2, keeping the `test-requirements` file in a separate commit.

imlopes (approved):
LGTM

connector/tests/test_advisory_lock.py
```diff
@@ -8,22 +8,26 @@
...
 from odoo.addons.connector.database import pg_try_advisory_lock
 from odoo.addons.queue_job.exception import RetryableJobError
 
 
-class TestAdvisoryLock(TransactionComponentCase):
+class TestAdvisoryLock(TransactionComponentRegistryCase):
     def setUp(self):
         super().setUp()
+        self._setup_registry(self)
+        self.comp_registry.load_components("component_event")
```
simahawk, line 20: why do we need this change? The `TransactionComponentRegistryCase` is needed only when you want to register custom components on the fly. When you load components from a module, `TransactionComponentCase` is enough

test-requirements.txt
```diff
@@ -0,0 +1,3 @@
+odoo-addon-component @ git+https://github.com/OCA/connector.git@refs/pull/521/head#subdirectory=component
```
simahawk, line 1: component and component_event are done

test-requirements.txt
```diff
@@ -0,0 +1,3 @@
...
+odoo-addon-component_event @ git+https://github.com/OCA/connector.git@refs/pull/509/head#subdirectory=component_event
+odoo-addon-queue_job==19.0.*
```
simahawk, line 3: queue job is available, this is not needed

simahawk (changes requested):

yankinmax:
Hello @mkoeck  if you don't mind, I've opened a new PR (to speed up migration process and address the suggestions) (your commits are preserved):

- [ ] https://github.com/OCA/connector/pull/531


@OCA/connector-maintainers  can you pls close this one?

mkoeck:
@yankinmax yes of course, thank you for taking over. Sorry for not being responsive, I am currently quite busy with work related issues, so I did not find the time to continue here

PR #512, opened 2025-12-19
draft: [16.0][ADD] connector_amazon_spapi

dnplkndll:


kobros-tech:
Product Bindings raised this error:

image

clementmbr:
Hi @dnplkndll !

FYI I started an Amazon connector in 2023-2024 with another strategy : 
https://github.com/akretion/sale-import/pull/83

I never finished it because the clients didn't have the budget anymore, but the idea was to be based on the [OCA/sale-channel](https://github.com/OCA/sale-channel/tree/16.0) modules and the [python-amazon-sp-api](https://github.com/saleweaver/python-amazon-sp-api/tree/master) library instead of rewriting all the connection and importation code in one big module.

(my PR is a bit messy because it is based on a akretion/sale-import repo and these base modules are now on OCA/sale-channel repo... but I hope you get the idea :) ) 

Another difference I notice is that you don't make the difference between **Seller** (selling to customers through Amazon) and **Vendor** (selling to Amazon) as Amazon does. It seems important to incorporate this logic in 3 different modules (base/vendor/seller).

Finally, I notice that you chose to recreate the whole sale.order(.line) logic with this amazon.sale.order(.line) model to handle the imported order's data before creating the real sale.order, while we used the "[sale.import.payload](https://github.com/OCA/sale-channel/blob/16.0/sale_import_base/models/sale_import_payload.py#L26)" model, which only contains the raw payload data as a big string, before creating the Odoo sale.order.

(in my messy PR I use the "[queue.job.chunk](https://github.com/akretion/sale-import/blob/16.0-add-sale-import-amazon/queue_job_chunk/models/queue_job_chunk.py#L24)" model, which was part of a separated module, but which is now included in the [sale_import_base](https://github.com/OCA/sale-channel/tree/16.0/sale_import_base) module, and renamed "sale.import.payload").

I admit this approach might be less user friendly, but it seems more robust... and might save more than 700 lines of code. 

what do you think?

cc @sebastienbeau @florian-dacosta @PaulGoubert

jelenapoblet:
Thanks for sharing, that’s useful context 👍
Rather than going too deep into architecture in the PR, what do you think about continuing this discussion on Discord, so we can first align on scope and structure? I'm pinging out to @etobella to suggest creating a dedicated connectors channel if that makes sense.
From our side, we do think a modular approach (instead of a single monolithic addon) is important from the start, especially with Odoo v18+ in mind.

dnplkndll:
we have a bit of spaghetti already . the have to do all the PRs on all the other projects get approved. deal with 100+ PR in gitagg in the mean time. mixed feelings on the subject but would love to discuss. I do not currently have the use case to test the FBA but ready to fo on amazon testing and was hoping to consolidate the missing parts like the pull or push competitor price match since the comparables live in seller central. then managing this as one of 3 channels I need to presever stock in total to customer in one of 5 sites in addition to market here. but other customer buying in regular versus sell on amazon until someone else gets the buy box is not the goal.
Another issue is I need to route and ship quote ( find the warehouse and the carrier ) I hopes to find an interface to do that or some sane defaults.

clementmbr:
Hi @jelenapoblet! Ok to be part of the OCA's Discord channel, my username is clementmbr there too.

I just won't have a lot of time to work on the topic, but happy to share what I've made and understood so far.

jelenapoblet:
Here's the link to the discord channel: https://discord.gg/GDz3kPse

kobros-tech:
@dnplkndll 

All GET requests are tested, no POST request is tested yet, but I am willing to get a volunteer to sacrifice himself for others to survive :)

kobros-tech:
@dnplkndll 

we can make it open now, let us give people here to review.

dnplkndll:
Closing in favor of a new PR with renamed module (`connector_amazon`) and `amz.*` model namespace to avoid conflicts with Odoo Enterprise `sale_amazon` (`amazon.*` namespace).

PR #513, opened 2025-12-28
draft: [16.0][ADD] connector_amazon_spapi 2

kobros-tech:
using python-amazon-sp-api


kobros-tech:
@jelenapoblet
@clementmbr

here is an update using third party connector

PR #524, opened 2026-03-26
[18.0] component: ease ComponentMixin setup

simahawk:
Split the part to make the registry ready so that depending module can set up the registry properly w/o relying on nested setUpcalls which might be not working properly in case of multiple inheritance.

Side effect of the Odoo tests apocalypse :/


lmignon (approved):

simahawk:
Tested :heavy_check_mark:  Fast tracking :speedboat: 
/ocabot merge patch

PR #530, opened 2026-05-07
Revert "component: allow ctx override at init"

simahawk:
This reverts commit ee7bd1a16ab8aa3af9144d042080445ad8b992e4 as it's breaking tests on depending modules.

From #504 

Unfortunately this works only w/ py > 3.10.

Explanation on #529


simahawk:
fast tracking
/ocabot merge patch

guewen (approved):

